PR #18, opened 2023-08-31
fix: fix error shown in debug message

warber:
Fixes a debug message:

debug   [coord=project:bucket:my-bucket][gid=0] Failed to create new object with bucket name "project_bucket_my-bucket". Trying to update existing object. Error: **%!s(<nil>)**

### DIFF
--- a/api/clients/buckets/bucket.go
+++ b/api/clients/buckets/bucket.go
@@ -290,7 +290,7 @@ func (c Client) upsert(ctx context.Context, bucketName string, data []byte) (Res
 	}
 
 	// Otherwise, try to update an existing bucket definition
-	logger.V(1).Info(fmt.Sprintf("Failed to create new object with bucket name %q. Trying to update existing object. Error: %s", bucketName, err))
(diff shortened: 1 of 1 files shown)


api/clients/buckets/bucket.go
```diff
@@ -290,7 +290,7 @@ func (c Client) upsert(ctx context.Context, bucketName string, data []byte) (Res
 	}
 
 	// Otherwise, try to update an existing bucket definition
-	logger.V(1).Info(fmt.Sprintf("Failed to create new object with bucket name %q. Trying to update existing object. Error: %s", bucketName, err))
+	logger.V(1).Info(fmt.Sprintf("Failed to create new object with bucket name %q. Trying to update existing object. Error: %s", bucketName, resp.Payload))
```
UnseenWizzard, line 293:
```suggestion
	logger.V(1).Info(fmt.Sprintf("Failed to create new object with bucket name %q. Trying to update existing object. API Error (HTTP %d): %s", bucketName, resp.StatusCode, resp.Payload))
```

I think it might be work making it clear it was an API error and include the status code

UnseenWizzard (approved):

PR #19, opened 2023-08-31
feat: add WithUserAgent to factory

warber:
Adds the possibility to set an "User-Agent" via the provided client factory.
Also, the struct name was changed to just `factory` instead of `clientFactory`
Further, accessing the factory is now just possible via an exported `Factory()` function

### DIFF
--- a/api/clients/factory.go
+++ b/api/clients/factory.go
@@ -42,6 +42,7 @@ func Factory() factory {
 type factory struct {
 	url         string                    // The base URL of the API.
 	oauthConfig *clientcredentials.Config // Configuration for OAuth2 client credentials.
+	userAgent   string                    // The User-Agent header to be set
 }
 
 // WithOAuthCredentials sets the OAuth2 client credentials configuration for the factory.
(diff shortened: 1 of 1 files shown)


api/clients/factory.go
```diff
@@ -26,35 +26,44 @@ import (
...
-func (f clientFactory) WithEnvironmentURL(u string) clientFactory {
+func (f factory) WithEnvironmentURL(u string) factory {
 	f.url = u
 	return f
 }
 
-// BucketClient creates and returns a new instance of buckets.Client for interacting with the bucket API
-func (f clientFactory) BucketClient() (*buckets.Client, error) {
+// WithUserAgent sets the User-Agent header.
+func (f factory) WithUserAgent(userAgent string) factory {
```
UnseenWizzard, line 60: Shouldn't we just allow setting arbitrary headers?

UnseenWizzard: Never mind, I guess in general headers are something you might want to set specifically for the client with SetHeader, while the user-agent is something you want on all clients

UnseenWizzard (approved):

api/clients/factory.go
```diff
@@ -26,35 +26,44 @@ import (
...
+	f.userAgent = userAgent
+	return f
+}
+
+// BucketClient creates and returns a new instance of buckets.Client for interacting with the bucket API.
+func (f factory) BucketClient() (*buckets.Client, error) {
```
UnseenWizzard, line 66: not necessary in this pr but in general I think it would make sense to unit test the behaviour of these client constructors. Does produce errors as expected, does it create a correctly configured bucket client, etc. This being the main interface I think it's well worth covering its expected behavior with tests

warber: tests are coming :)

PR #23, opened 2023-09-01
feat: add possibility to convert an Response into an APIError

warber:
Often you want to treat API responses (4xx,5xx) as a real go error. I've introduced the AsAPIError method on response which enables just that.

```go
if err, ok := response.AsAPIError(); ok {
  //do something wit api err
}
```

### DIFF
--- a/api/response_test.go
+++ b/api/response_test.go
@@ -15,7 +15,9 @@
 package api
 
 import (
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
 	"github.com/stretchr/testify/assert"
+	"net/http"
 	"testing"
 )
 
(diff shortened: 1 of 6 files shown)


api/response_test.go
```diff
@@ -97,3 +99,72 @@ func TestDecodeJSONObjects(t *testing.T) {
...
+					URL:    "https://www.dt.com/resources",
+				},
+			}
+
+			err, ok := resp.AsAPIError()
+
+			assert.IsType(t, tc.expected, err)
+			assert.Equal(t, tc.expected, err)
+
+			if tc.shouldFail {
```
UnseenWizzard, line 163: this field doesn' make sense, neither in naming nor usage. I'll assume this is another reason to not trust AI double-check generated code ;) 

an `expectedOK` and simple `assert.Equal(t, tc.expectedOK, ok)` should do the trick

api/rest/client_test.go
```diff
@@ -72,27 +72,31 @@ func TestClient_CRUD(t *testing.T) {
 
 		assert.NoError(t, err, "Unexpected error")
 		assert.Equal(t, http.StatusOK, resp.StatusCode, "Expected status code 200")
+		assert.Equal(t, http.MethodGet, resp.RequestInfo.Method)
```
UnseenWizzard, line 75: if we test that the method is set, shouldn't we the assert the URL string as well?

UnseenWizzard (changes requested):
please fix the response_test setup, it's misleading at the moment

UnseenWizzard (approved):

PR #25, opened 2023-09-04
Fix/bucket updates

UnseenWizzard:

- [fix(buckets): Only attempt update when upserting if create failed with a conflict](https://github.com/Dynatrace/dynatrace-configuration-as-code-core/pull/25/commits/3f504a43a3cadc8113b7e51fe74a04d718319714)
- [feat(buckets): Do not update buckets if they are unmodified](https://github.com/Dynatrace/dynatrace-configuration-as-code-core/pull/25/commits/0b4aff799f4164c0e890847f796afcce3d06e1e5)

- An update is only attempted on Upsert if creation failed with a conflict, any other errors like permissions or bad payloads are returned immediately, and no longer hidden behind a likely subsequent GET error when trying to update.
- Upsert and Update no longer update an object if the existing values are the same as what they would be updated to. From an interface perspective, nothing changes and users are still supplied with a 200 Response with empty payload, same as what the Dynatrace API returns.

### DIFF
--- a/api/clients/buckets/bucket.go
+++ b/api/clients/buckets/bucket.go
@@ -55,8 +55,27 @@ type listResponse struct {
 	Buckets []json.RawMessage `json:"buckets"`
 }
 
+type retrySettings struct {
+	maxRetries   int
+	waitDuration time.Duration
+}
+
 type Client struct {
-	client *rest.Client
+	client        *rest.Client
+	retrySettings retrySettings
+}
+
+// Option represents a functional Option for the Client.
+type Option func(*Client)
+
+// WithUpdateRetrySettings sets the maximum number of retries as well as duration between retries for Update and Upsert HTTP requests
+func WithUpdateRetrySettings(maxRetries int, waitDuration time.Duration) Option {
+	return func(c *Client) {
+		c.retrySettings = retrySettings{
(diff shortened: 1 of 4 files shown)


api/clients/buckets/bucket.go
```diff
@@ -334,6 +345,8 @@ func (c Client) list(ctx context.Context) (listResponse, error) {
 	return l, nil
 }
 
+var unmodified = errors.New("unmodified bucket")
```
UnseenWizzard, line 348: I'm not sure if a sentinel error is the best way to handle this, but I honestly couldn't think of a decent alternative that allows the calling method to discern this case.
It needs to know, as otherwise this case is handled and logged the same as a successful HTTP PUT update.

UnseenWizzard, line 348: I'm not sure if a sentinel error is the best way to handle this, but I honestly couldn't think of a decent alternative that allows the calling method to discern this case.
It needs to know, as otherwise this case is handled and logged the same as a successful HTTP PUT update.

api/clients/buckets/bucket.go
warber, line 1: pls update code docs ( especially of Update()/Upsert())

warber (approved):

PR #27, opened 2023-09-05
Refactor: Simplify and refactor `RequestResponseRecorder`

warber:
The `RequestResponseRecorder` component was a bit clumsy to use and a bit over-engineered.
Instead of a channel based approach (which is actually really not needed as we anyway want to have the full synchroneous behavior and are just simulating it with a non-buffered channel), I've followd a much simpler "callback-based" approach.

The component is also renamed to "HTTPListener" because that's what it is essentially.
Further, the id is now generated in the client. Passed as part of the RequestResponse struct.

I thought about doing the generation of the ID inside the Listener component, so that the caller does not need to know to
pass the **same** id to the methods. But this just introduced unnecessary noice and complexity. so I've just introduced an "id" argument.


UnseenWizzard (approved):
lgtm

PR #28, opened 2023-09-05
Doc: update `README.md` with example showing how to use HTTPListener

warber:
Update `README.md` to describe newest features of lib.


README.md
```diff
@@ -78,6 +81,25 @@ ctx := logr.NewContext(context.TODO(), logrusr.New(logrus.New()))
 resp, err := ctx.Get(ctx,"...")
 ```
 
+### Tracking and logging HTTP requests/responses
+If you want to keep track or just log all HTTP requests/responses that are happening as part of the execution of the clients you can attach an `HTTPListener` that can be used
```
UnseenWizzard, line 85:
```suggestion
If you want to keep track or just log all HTTP requests/responses happening as part of the execution of the clients, you can implement an `HTTPListener` and attach it to the client.
```

UnseenWizzard (approved):

PR #34, opened 2023-09-19
feat: introduce api client for automation resources

warber:
This PR introduces a new client targeting the automation API. It provides support for workflows, business-calendars, and scheduling rules.

The client can be obtained via the factory as usual and offers CRUD + Upsert operations.
I have not yet added logging to the client.
This is planned for a follow-up PR to solely focus on useful logs.

Yes, user has access to new automation client, e.g. via the existing factory:

```go
client, err := factory.AutomationClient()
if err != nil {
   // handle err
}
```

from here on you can call the desired operation on the client, e.g. getting an existing workflow:
```go
response, err = automationClient.Get(context.TODO(), automation.Workflows, id)
if err != nil {
	panic(err)
}

if err, isAPIErr := response.AsAPIError(); isAPIErr {
	panic(err)
}
....
```

### DIFF
--- a/internal/testutils/testserver.go
+++ b/internal/testutils/testserver.go
@@ -24,6 +24,7 @@ import (
 
 // TestServer is a wrapper around httptest.Server that provides utility methods for testing.
 type TestServer struct {
+	calls            int
 	*httptest.Server // Embedded httptest.Server for underlying server functionality.
 }
 
@@ -31,6 +32,11 @@ type TestServer struct {
 func (t TestServer) URL() *url.URL {
 	u, _ := url.Parse(t.Server.URL) //nolint:errcheck
 	return u
+}
+
+// Calls returns the number of calls invoked on the test server
+func (t TestServer) Calls() int {
+	return t.calls
 }
 
 // Client returns an HTTP client associated with the test server.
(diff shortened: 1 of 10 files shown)


internal/testutils/testserver.go
```diff
@@ -56,9 +62,31 @@ type ServerResponses map[HTTPMethod]struct {
...
-func NewHTTPTestServer(t *testing.T, arg ServerResponses) *TestServer {
-	return &TestServer{Server: httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		if res, found := arg[req.Method]; found {
+func NewHTTPTestServer(t *testing.T, responses []ServerResponses) *TestServer {
+	testServer := &TestServer{}
+	handler := func(rw http.ResponseWriter, req *http.Request) {
+		testServer.calls++
+		// Special case: if only one response is specified, this will be reused
+		// for each and every call
+		if len(responses) == 1 {
```
warber, line 71: I don't like that implicit behavior but it is useful sometimes and already existing tests are relying on that

internal/testutils/testserver.go
```diff
@@ -56,9 +62,31 @@ type ServerResponses map[HTTPMethod]struct {
...
+				_, _ = rw.Write([]byte(res.ResponseBody)) // nosemgrep: go.lang.security.audit.xss.no-direct-write-to-responsewriter.no-direct-write-to-responsewriter
+			} else {
+				t.Errorf("unexpected HTTP method call: %s", req.Method)
+			}
+			return
+		}
+
+		if len(responses) <= testServer.calls-1 {
+			t.Fail()
+			t.Fatalf("Exceeded number of calls to test server (expected: %d)", len(responses))
```
UnseenWizzard, line 86: I think `Fatal/Fatalf` already call a `FailNow`(no guarantee), so the t.Fail() on line 85 isn't needed

internal/testutils/testserver.go
UnseenWizzard: That seems pretty strange to use though? And gives me no way to assert that a test server receives exactly one API call, rather than many...

I don't know which tests rely on this behavior but IMO it'd be better to re-write them to define their responses and expected requests explicitly. 

Awesome to see that you introduced this idea btw!

internal/testutils/testserver.go
UnseenWizzard: Also general note: I think this server should be consistent on when it stops a test. 
As is an unexpected HTTP method will fail the test, but let it continue (Errorf), but a totally unexpected call will fail the test immediately (Fatalf). 

I don't have a strong preference (generally I prefer stopping, but failing and continuing seems to be the more usual default), but I think this should use Errorf or Fatalf exclusively.

UnseenWizzard (approved):
Overall looks good to me :) Comments are mostly improvement suggestions

UnseenWizzard (approved):

PR #36, opened 2023-09-20
chore: enhance `testutils.TestServer`

warber:
Use dedicated struct with functions instead of raw map.

### DIFF
--- a/api/clients/automation/automation_test.go
+++ b/api/clients/automation/automation_test.go
@@ -36,8 +36,8 @@ func TestAutomationClient_Get(t *testing.T) {
 	const payload = `{ "id" : "91cc8988-2223-404a-a3f5-5f1a839ecd45", "data" : "some-data1" }`
 
 	t.Run("Get - no ID given", func(t *testing.T) {
-		responses := testutils.ServerResponses{}
-		server := testutils.NewHTTPTestServer(t, []testutils.ServerResponses{responses})
+		responses := []testutils.ResponseDef{}
+		server := testutils.NewHTTPTestServer(t, responses)
 		defer server.Close()
 
 		client := automation.NewClient(rest.NewClient(server.URL(), server.Client()))
@@ -49,13 +49,18 @@ func TestAutomationClient_Get(t *testing.T) {
 
 	t.Run("GET - OK", func(t *testing.T) {
 
-		responses := testutils.ServerResponses{
-			http.MethodGet: {
-				ResponseCode: http.StatusOK,
-				ResponseBody: payload,
+		responses := []testutils.ResponseDef{
+			{
+				GET: func(t *testing.T, req *http.Request) testutils.Response {
+					return testutils.Response{
+						ResponseCode: http.StatusOK,
+						ResponseBody: payload,
+					}
+				},
 			},
 		}
-		server := testutils.NewHTTPTestServer(t, []testutils.ServerResponses{responses})
+
+		server := testutils.NewHTTPTestServer(t, responses)
 		defer server.Close()
 
 		client := automation.NewClient(rest.NewClient(server.URL(), server.Client()))
@@ -68,8 +73,9 @@ func TestAutomationClient_Get(t *testing.T) {
 	})
 
 	t.Run("GET - Unable to make HTTP call", func(t *testing.T) {
-		responses := testutils.ServerResponses{}
-		server := testutils.NewHTTPTestServer(t, []testutils.ServerResponses{responses})
+
+		responses := []testutils.ResponseDef{}
+		server := testutils.NewHTTPTestServer(t, responses)
 		defer server.Close()
 
 		client := automation.NewClient(rest.NewClient(server.URL(), server.FaultyClient()))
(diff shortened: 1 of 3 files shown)


api/clients/automation/automation_test.go
```diff
@@ -162,28 +187,34 @@ func TestAutomationClient_Update(t *testing.T) {
...
-					adminAccessQP := request.URL.Query()["adminAccess"]
+		responses := []testutils.ResponseDef{
+			{
+				PUT: func(t *testing.T, req *http.Request) testutils.Response {
+					return testutils.Response{
+						ResponseCode: http.StatusForbidden,
+						ResponseBody: payload,
+					}
+				},
+				ValidateRequest: func(t *testing.T, req *http.Request) {
```
UnseenWizzard, line 198: nit/thought: With the responses now being functions that get the request as well as testing, I'm starting to wonder if having a dedicated Validate function is even required. I could just do the asserts as part of the response function - in the same way I would when writing a httptest.HandlerFunc.
What do you think?

warber: I left it separated, thinking further test utilities might provide some sort of "default" validators you could just plug in.

internal/testutils/testserver.go
```diff
@@ -51,55 +51,87 @@ func (t TestServer) FaultyClient() *http.Client {
...
-			_, _ = rw.Write([]byte(res.ResponseBody)) // nosemgrep: go.lang.security.audit.xss.no-direct-write-to-responsewriter.no-direct-write-to-responsewriter
-		} else {
-			t.Errorf("unexpected HTTP method call: %s", req.Method)
+		handlerFunc, found := methodFuncs[req.Method]
+		if !found {
+			rw.WriteHeader(http.StatusMethodNotAllowed)
+			_, _ = rw.Write([]byte("Method not supported")) // nosemgrep: go.lang.security.audit.xss.no-direct-write-to-responsewriter.no-direct-write-to-responsewriter
+			return
 		}
+		if handlerFunc == nil {
```
UnseenWizzard, line 77: isn't that other than a strange edge-case the same as `!found` ? 
Why handle no response function being found as a HTTP error, and explicitly configuring nil for a method as a test error?

internal/testutils/testserver.go
```diff
@@ -51,55 +51,87 @@ func (t TestServer) FaultyClient() *http.Client {
...
 	return testServer
 }
 
+type Response struct {
+	ResponseCode int
+	ResponseBody string
+}
+
+type ResponseDef struct {
+	GET             func(*testing.T, *http.Request) Response
```
UnseenWizzard, line 97: nit: It might be a bit confusing that you can define several methods for a single response - and the server will answer to whatever was requested. Can't think of a better way than going back to a map[string]func though...

internal/testutils/testserver.go
warber: :D yes, should we treat `!found` just as a test error?

internal/testutils/testserver.go
```diff
@@ -51,55 +51,87 @@ func (t TestServer) FaultyClient() *http.Client {
...
+	PUT             func(*testing.T, *http.Request) Response
+	POST            func(*testing.T, *http.Request) Response
+	DELETE          func(*testing.T, *http.Request) Response
+	ValidateRequest func(*testing.T, *http.Request)
+}
+
+func (r ResponseDef) Get(t *testing.T, req *http.Request) Response {
```
UnseenWizzard, line 104: are these used anywhere? if I'm not mistaken you only use the struct fields directly?

UnseenWizzard (commented):
Nice change! Might benefit from a bit of cleanup still.

internal/testutils/testserver.go
warber: I thought of including a check that panics if you define more than one, wdyt?

internal/testutils/testserver.go
UnseenWizzard: I'd say it makes sense, yes. 
If I didn't define that the server should reply to a PUT, and it gets an expected one, it would be reasonable to see that as a failing test IMO

internal/testutils/testserver.go
warber: only `Validate` is used which provides a default passthrough if you didn't specify a ValidateRequest func.
forgot to call the others as well.

internal/testutils/testserver.go
UnseenWizzard: That'd feel in line with what we said above about failing if unexpected calls happen. I don't know if there'd ever be a case, if I want to define an either/or kind of reply for a test... 🤔 Probably not. 
Maybe t.Fatal would be cleaner than just panicing?

internal/testutils/testserver.go
UnseenWizzard: Do you want to call them or remove them, though? 
With the current map and !found behavior there can't be a case that needs the != nil check these methods provide.
Could replace the map with a switch and method call maybe, but I don't know if that improves anything

internal/testutils/testserver.go
warber: not sure, actually panic makes the most sense to me.
actually also in the case where a handler is `!found` it should panic.
It's no test error, it's just wrongly used and could never work.

That's also how https://github.com/matryer/moq are handling these situations. if something not defined on your mock is called, it panics.

are you fine with that?

internal/testutils/testserver.go
warber: I would call them, it gives the possibility to add additional logic if needed (maybe never will be but it does not hurt if they are there, and for Validate it is actually already used)

warber:
@UnseenWizzard I've addressed the issues. For now, I've decided to let the testing server `panic` if anything related to testsetup (e.g. missing handler function or more than one handler function defined,...) occurs. I could not see any particular benefit of usigng e.g. `t.Fatal` tho I am relaxed about changing it to this if you insist :)

internal/testutils/testserver.go
```diff
@@ -51,59 +52,108 @@ func (t TestServer) FaultyClient() *http.Client {
...
+
 // ErrorTransport is custom transport that always produces a simulated network error.
 type ErrorTransport struct{}
 
 // RoundTrip implements the RoundTripper interface and returns a simulated network error.
 func (t *ErrorTransport) RoundTrip(_ *http.Request) (*http.Response, error) {
 	return nil, errors.New("simulated network error")
 }
+
+func checkExactlyOneHandlerSet(def ResponseDef) bool {
```
UnseenWizzard, line 144: nit: technically the name is incorrect. It allows zero handlers as well - not sure if it even should though? 
If I want to assert that no API calls happen, a testserver with zero responses would also make the test fail if it gets one, right?

UnseenWizzard (dismissed):

PR #40, opened 2023-09-29
refactor: unify return types of clients

warber:
Although achieving uniformity across all clients' interfaces may not be practical, standardizing the return signature can be easily accomplished. This simplifies the usage of the client(s) and enhances convenience.
Moreover, these changes support the flexibility to not break any code once an API suddenly introduces pagination.

:exclamation: Return types are different, thus clients need to adopt the usage.

### DIFF
--- a/api/clients/automation/automation.go
+++ b/api/clients/automation/automation.go
@@ -28,9 +28,7 @@ import (
 	"strconv"
 )
 
-type Response struct {
-	api.Response
-}
+type Response = api.Response
 
 type ListResponse = api.PagedListResponse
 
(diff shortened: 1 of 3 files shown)


api/clients/automation/automation.go
```diff
@@ -28,9 +28,7 @@ import (
 	"strconv"
 )
 
-type Response struct {
-	api.Response
-}
+type Response = api.Response
```
jskelin, line 31: Is it possible to remove the alias also?
I don't see added value

api/clients/buckets/bucket.go
```diff
@@ -32,22 +32,18 @@ import (
...
-}
-
 type response struct {
 	api.Response
 	BucketName string `json:"bucketName"`
 	Status     string `json:"status"`
 	Version    int    `json:"version"`
 }
 
+type Response = api.Response
```
jskelin, line 42: Is it possible to remove the alias also?
I don't see added value

api/clients/buckets/bucket.go
```diff
@@ -32,22 +32,18 @@ import (
...
+
 // ListResponse is a Bucket API response containing multiple bucket objects.
 // For convenience, it contains a slice of Buckets in addition to the base api.Response data.
-type ListResponse struct {
-	api.ListResponse
-}
+type ListResponse = api.PagedListResponse
```
jskelin, line 46: Is it possible to remove the alias also?
I don't see added value

jskelin (approved):

api/clients/buckets/bucket.go
warber: The advantage is, that you don't need to import two packages but just e.g. "github.com/dynatrace/dynatrace-configuration-as-code-core/api/clients/buckets" when using that client.

Laubi (approved):

PR #47, opened 2023-10-17
feat(buckets): Backoff when retrying upserts and retry creation if a bucket is being deleted

UnseenWizzard:


api/clients/buckets/bucket.go
```diff
@@ -194,6 +204,9 @@ func (c Client) Create(ctx context.Context, bucketName string, data []byte) (Res
 	}, nil
 }
 
+var DeletingBucketErr = errors.New("cannot update bucket that is currently being deleted")
+var UpdateStepFailed = errors.New("failed ")
```
UnseenWizzard, line 208: - [ ] @UnseenWizzard remove, not used

clients/buckets/bucket.go
```diff
@@ -0,0 +1,569 @@
+// @license
```
UnseenWizzard, line 1: Note: This is functionally the same `api/client/buckets` client as before, but direct calls to the rest.Client are replaced with using the newly introduced "plain API CRUD" bucket client.

(automation for now is just moved without introducing the plain API client, so we can review the concept on one sample)

api/clients/buckets/bucket.go
```diff
@@ -100,469 +39,63 @@ func WithRetrySettings(maxRetries int, durationBetweenTries time.Duration, maxWa
...
-	}
-
-	for _, o := range option {
-		o(c)
 	}
 
 	return c
 }
 
-// Get retrieves a bucket definition based on the provided bucketName. The function sends a GET request
```
UnseenWizzard, line 112: Note: I've omitted documentation for now, this initial change is just so we can review the concept.

api/clients/buckets/bucket.go
```diff
@@ -100,469 +39,63 @@ func WithRetrySettings(maxRetries int, durationBetweenTries time.Duration, maxWa
...
-		if _, err := c.awaitBucketState(ctx, bucketName, removed); err != nil {
-			return Response{}, err
-		}
-		return c.Create(ctx, bucketName, data)
-	}
-	return apiResp, err
-}
-
-func (c Client) create(ctx context.Context, bucketName string, data []byte) (rest.Response, error) {
+func (c Client) Create(ctx context.Context, bucketName string, data []byte) (rest.Response, error) {
```
UnseenWizzard, line 52: The biggest question I have about this level of client is, what types should they return?

For now, I've stuck to `rest.Response`, as `api.Response` already needs some logic to be constructed.
This is not noticeable here, but will be for automation -> IMO a plain API client for automation should not resolve pagination and thus just return one rest.Response which might contain info about additional pages.

If these clients shouldn't use `api.Response` though, it should probably move up to the new `client/` package level too.

clients/buckets/bucket.go
```diff
@@ -0,0 +1,569 @@
...
+	resp, err := c.apiClient.Delete(ctx, bucketName)
+	if err != nil {
+		return api.Response{}, fmt.Errorf("unable to delete object with bucket name %q: %w", bucketName, err)
+	}
+	if !resp.IsSuccess() {
+		return api.Response{StatusCode: resp.StatusCode, Data: resp.Payload, Request: resp.RequestInfo}, nil
+	}
+
+	// await bucket being successfully deleted
+	timoutCtx, cancel := context.WithTimeout(ctx, c.retrySettings.maxWaitDuration)
```
Laubi, line 340:
```suggestion
	timeoutCtx, cancel := context.WithTimeout(ctx, c.retrySettings.maxWaitDuration)
```

Laubi (approved):

Laubi (approved):

PR #61, opened 2023-12-20
fix: make GetGroupDto.Uuid a required value (not a pointer)

jskelin:
Generated code made `GetGroupDto.Uuid` a pointer although by the OpenAPI definition, this field is always present. This change will make UUID not a pointer.

Does any work need to be done in the generator?


warber:
@jskelin IMO, making manual changes to generated code is very bad practice. Hence I would rather propose to either adapt the template or (preferably) the json schema.

jskelin:
until team decision on design aborted

PR #89, opened 2024-03-25
Retry after admin permission reject for ALL resources

Dynatrace-Reinhard-Pilz:
Automation Clients are sending the query parameter `adminAccess` - attempting to gain admin permissions in order to manage not just the settings owned by the user, but all of them. That parameter is getting sent along for all three resources within the Automation API - Workflows, Business Calendars and Scheduling Rules.

If that user isn't eligible to get these elevated permissions, the retry is attempted without `adminAccess` - but currently that retry happens for Workflow requests **only**. 

This PR ensures that the retry without `adminAccess` parameter is being attempted for **all** resources.


api/clients/automation/automation.go
```diff
@@ -202,19 +202,19 @@ func (a Client) Delete(ctx context.Context, resourceType ResourceType, id string
 }
 
 func (a Client) makeRequestWithAdminAccess(resourceType ResourceType, request func(options rest.RequestOptions) (*http.Response, error)) (*http.Response, error) {
```
UnseenWizzard, line 204: nit/later refactoring suggestion: as this method is called `makeRequestWithAdminAccess`, wouldn't it be cleaner to have the `if resourceType == Workflows` check outside before we call the method? If workflow -> withAdminAcces, else -> request()

As this won't functionally change the fix or the core libraries interfaces, fine to merge as is and follow up later

UnseenWizzard (approved):

PR #93, opened 2024-04-18
chore: Update bucket client to return API errors

arthurpitman:
**What this PR does / Why we need it:**
To align this client with the newly created document client (https://github.com/Dynatrace/dynatrace-configuration-as-code-core/pull/91), this PR changes the way the bucket client behaves in error cases.

Instead of always returning a response, it now returns an `APIErrorerror` type in case the API request resulted in a HTTP error status != 2xx.

Errors that occur during reading the body currently do not include a status code, however this is something we may address in a follow-up ticket.

### DIFF
--- a/clients/buckets/bucket.go
+++ b/clients/buckets/bucket.go
@@ -165,11 +165,11 @@ func (c Client) List(ctx context.Context) (ListResponse, error) {
 	if err != nil {
 		return ListResponse{}, fmt.Errorf("failed to list buckets:%w", err)
 	}
+	defer resp.Body.Close()
 
 	// if the response has code != 2xx return empty list with response info
 	if !rest.IsSuccess(resp) {
 		body, err := io.ReadAll(resp.Body)
-		resp.Body.Close()
 		if err != nil {
 			return ListResponse{}, err
 		}
@@ -217,8 +217,9 @@ func (c Client) Create(ctx context.Context, bucketName string, data []byte) (Res
 	if err != nil {
 		return api.Response{}, err
 	}
+	defer resp.Body.Close()
+
 	body, err := io.ReadAll(resp.Body)
-	resp.Body.Close()
 	if err != nil {
 		return Response{}, err
 	}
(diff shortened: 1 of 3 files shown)


clients/buckets/bucket.go
```diff
@@ -161,6 +165,18 @@ func (c Client) List(ctx context.Context) (ListResponse, error) {
 	if err != nil {
 		return ListResponse{}, fmt.Errorf("failed to list buckets:%w", err)
 	}
+
+	// if the response has code != 2xx return empty list with response info
+	if !rest.IsSuccess(resp) {
+		body, err := io.ReadAll(resp.Body)
+		resp.Body.Close()
+		if err != nil {
+			return ListResponse{}, err
```
UnseenWizzard, line 174: You're swallowing the actual response error in case we can't read the body. This will make it impossible for a user of the client to actually know that and how the REST API call has failed. I'd still return an API error without body and log that we failed to read the body

UnseenWizzard: Actually this is something that's now happening in several cases in this client. @warber is this also a pattern you've adopted in the other client changes? IMO that's problematic when using these clients. Reading the response body is not the main action of these methods, so we shouldn't just not return any information about the actual REST response if that fails.

clients/buckets/bucket.go
```diff
@@ -258,14 +274,8 @@ func (c Client) Update(ctx context.Context, bucketName string, data []byte) (Res
 		if err != nil {
 			return Response{}, err
```
UnseenWizzard, line 277: Actually what I question above, was already the case here before the changes..

clients/buckets/bucket.go
arthurpitman: That is very true: the original error is more important - I will update all of these 👍

arthurpitman: @warber, we should align all clients here

UnseenWizzard (approved):
as discussed we merge this then unify all new/changed clients to not lose API error information if io.Read fails when already trying to create an API error

PR #96, opened 2024-04-22
fix: Return API error on Body Read Failures

warber:


clients/automation/automation.go
```diff
@@ -31,6 +32,8 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
 )
 
+const bodyReadErrMsg = "unable to read api response body"
```
UnseenWizzard, line 35: nit
```suggestion
const bodyReadErrMsg = "unable to read API response body"
```

clients/automation/automation.go
```diff
@@ -218,20 +224,22 @@ func (a Client) List(ctx context.Context, resourceType automation.ResourceType)
...
-				return ListResponse{}, err
-			}
-
 			return ListResponse{}, api.NewAPIErrorFromResponseAndBody(resp, body)
 		}
 
 		result, err = unmarshalJSONList(resp)
 		if err != nil {
-			return ListResponse{}, fmt.Errorf("failed to parse list response:%w", err)
+			logr.FromContextOrDiscard(ctx).Error(err, "failed to unmarshal json response")
```
UnseenWizzard, line 241: I wonder if we should be returning this as an API error - that feels a bit weird for a 2xx status code.

clients/buckets/bucket.go
```diff
@@ -30,6 +30,8 @@ import (
 	"time"
 )
 
+const bodyReadErrMsg = "unable to read api response body"
```
UnseenWizzard, line 33: nit^2:
```suggestion
const bodyReadErrMsg = "unable to read API response body"
```

clients/automation/automation.go
warber: due to lack of significantly better alternatives, i think it is ok
returning `responseVal, err` requires different handling on the client side,
leaving it like it is, hides the http response code,
introducing another type of error just for that seems awkward....

PR #112, opened 2024-06-24
chore: removed editable option from openpipeline client methods

warber:
As we are not sure if that's needed, I've removed the possiblility to filter by editable openipeline configs.
With this, all of the option structs are empty -so I've also removed them from the function signatures.

### DIFF
--- a/clients/openpipeline/openpipeline.go
+++ b/clients/openpipeline/openpipeline.go
@@ -23,7 +23,6 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
 	"github.com/go-logr/logr"
 	"io"
-	"slices"
 )
 
 const bodyReadErrMsg = "unable to read API response body"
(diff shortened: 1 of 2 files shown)


clients/openpipeline/openpipeline.go
```diff
@@ -35,20 +34,6 @@ type ListResponse struct {
 	Editable bool   `json:"editable"`
```
arthurpitman, line 34: Just to clarify: we keep this such that users of the client could check easily if the config is editable without having to get each one?

warber: yes, maybe we are interested in that one

clients/openpipeline/openpipeline_test.go
```diff
@@ -217,7 +210,21 @@ func TestOpenPipelineClient_GetAll(t *testing.T) {
...
+	const payloadGet2 = `{
+	"id": "events",
+	"editable": true,
+	"version": "1716904550612-4770deb9105b4a5293c1edbcc6bf4412"
+}
+`
+
+	const payloadGet3 = `{
 	"id": "bizevents",
 	"editable": true,
```
arthurpitman, line 229:
```suggestion
	"editable": false,
```

arthurpitman (approved):

PR #118, opened 2024-07-17
Feat/documents/patch public

jskelin:
Since Dynatrace HTTP API doesn't provide a way to create a public document via the PUT call, it is necessary to perform another PATCH call. With these changes, this is done automatically in the document client. Additionally, if the creation (1. step) fails, rollback (DELETE) is performed

### DIFF
--- a/clients/documents/documents.go
+++ b/clients/documents/documents.go
@@ -313,9 +313,7 @@ func (c Client) Delete(ctx context.Context, id string) (api.Response, error) {
 		return api.Response{}, err
 	}
 
-	resp, err := c.client.Delete(ctx, id, rest.RequestOptions{
-		QueryParams: map[string][]string{optimisticLockingHeader: {fmt.Sprint(getResp.Version)}},
-	})
+	resp, err := c.client.Delete(ctx, id, getResp.Version)
 	if err != nil {
 		return api.Response{}, err
 	}
--- a/clients/documents/documents.go
+++ b/clients/documents/documents.go
@@ -20,9 +20,12 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"mime"
 	"mime/multipart"
+	"net/http"
 	"net/url"
 	"strings"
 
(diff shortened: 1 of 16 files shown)


clients/documents/documents_test.go
```diff
@@ -15,14 +15,20 @@
...
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
+	docApi "github.com/dynatrace/dynatrace-configuration-as-code-core/api/clients/documents"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/clients/documents"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/testutils"
 	"github.com/stretchr/testify/assert"
-	"net/http"
-	"testing"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
```
warber, line 31: why did you choose to introduce a mocking library and write tests in different style than the existing ones?

clients/documents/documents.go
```diff
@@ -204,239 +216,194 @@ func (c Client) List(ctx context.Context, filter string) (ListResponse, error) {
...
-	if !rest.IsSuccess(resp) {
-		return Response{}, api.APIError{
-			StatusCode: resp.StatusCode,
-			Body:       body.Bytes(),
-			Request:    rest.RequestInfo{Method: resp.Request.Method, URL: resp.Request.URL.String()},
-		}
+	d := documents.Document{
+		Kind:    string(documentType),
+		Name:    name,
+		Public:  !isPrivate,
```
warber, line 268: maybe we can change the field name/semantics to `Private`?

jskelin: If we agreed that private is the default, I would keep the name `Public` as in go false is the default value.

In case we when the default for the new document is public, I would go with `Private`

clients/documents/documents_test.go
jskelin: As this testing framework is more mature and widely used. And perfectly fit for this purpose.
E.g., in the custom framework, if some call is defined but never happens, the tests pass.

warber (approved):

PR #119, opened 2024-07-23
fix: make creation of public documents more robust

warber:
The PATCH endpoint is called immediately after POST when creating a document in order to make it public. However, it can be the case that the subsequent call to the PATCH endpoint happens to early. In this situation the API returns 404 and the call to Create fails. To make this more stable, a retry mechanism was added.

### DIFF
--- a/clients/documents/documents.go
+++ b/clients/documents/documents.go
@@ -28,6 +28,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/clients/documents"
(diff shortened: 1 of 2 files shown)


arthurpitman (approved):

clients/documents/documents.go
```diff
@@ -275,6 +277,24 @@ func (c Client) create(ctx context.Context, d documents.Document) (api.Response,
 	return processHttpResponse(c.client.Create(ctx, d))
 }
 
+func (c Client) patchWithRetry(ctx context.Context, id string, version int, d documents.Document) (resp api.Response, err error) {
```
UnseenWizzard, line 280: the underlying core client has a built-in retry mechanism AFAIK (at least it still calls a sendWithRetries method) - no need to reinvent it here, if we setup the options correctly. 

```go
func (c *Client) PATCH(ctx context.Context, endpoint string, body io.Reader, options RequestOptions) (*http.Response, error) {
	return c.sendRequestWithRetries(ctx, http.MethodPatch, endpoint, body, options)
}
```

UnseenWizzard: addition: if I'm reading it correctly, the higher level client is intialized with a *rest.Client -> so shouldn't the issue we've run into already be covered via retry options configured on that client? Or does Monaco not set any retry options for the rest client it initializes?

PR #126, opened 2024-08-20
feat: Support per call retry settings

arthurpitman:
This PR adds an optional `*RequestRetrier` to `RequestOptions`, allowing per call retry settings to be specified which override the general ones for the client.

In addition, `DelayAfterRetry` is added to `RequestRetrier`, allowing the delay between attempts to also be varied for each call or the client the general.

### DIFF
--- a/api/rest/client.go
+++ b/api/rest/client.go
@@ -236,8 +236,8 @@ func (c *Client) sendWithRetries(ctx context.Context, req *http.Request, retryCo
 
 	if requestRetrier != nil && retryCount < requestRetrier.MaxRetries &&
 		requestRetrier.ShouldRetryFunc != nil && requestRetrier.ShouldRetryFunc(response) {
-		logger.V(1).Info(fmt.Sprintf("Retrying failed request %q (HTTP %s) after %d ms delay... (try %d/%d)", req.URL, response.Status, 100, retryCount+1, requestRetrier.MaxRetries), "statusCode", response.StatusCode, "try", retryCount+1, "maxRetries", requestRetrier.MaxRetries)
-		time.Sleep(100 * time.Millisecond)
+		logger.V(1).Info(fmt.Sprintf("Retrying failed request %q (HTTP %s) after %d ms delay... (try %d/%d)", req.URL, response.Status, requestRetrier.DelayAfterRetry.Milliseconds(), retryCount+1, requestRetrier.MaxRetries), "statusCode", response.StatusCode, "try", retryCount+1, "maxRetries", requestRetrier.MaxRetries)
+		time.Sleep(requestRetrier.DelayAfterRetry)
 		return c.sendWithRetries(ctx, req, retryCount+1, options)
 	}
 	return response, nil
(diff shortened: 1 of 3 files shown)


api/rest/client.go
```diff
@@ -225,10 +234,10 @@ func (c *Client) sendWithRetries(ctx context.Context, req *http.Request, retryCo
 		c.rateLimiter.Update(ctx, response.StatusCode, response.Header)
 	}
 
-	if c.requestRetrier != nil && retryCount < c.requestRetrier.MaxRetries &&
-		c.requestRetrier.ShouldRetryFunc != nil && c.requestRetrier.ShouldRetryFunc(response) {
-		logger.V(1).Info(fmt.Sprintf("Retrying failed request %q (HTTP %s) after %d ms delay... (try %d/%d)", req.URL, response.Status, 100, retryCount+1, c.requestRetrier.MaxRetries), "statusCode", response.StatusCode, "try", retryCount+1, "maxRetries", c.requestRetrier.MaxRetries)
-		time.Sleep(100 * time.Millisecond)
+	if requestRetrier != nil && retryCount < requestRetrier.MaxRetries &&
+		requestRetrier.ShouldRetryFunc != nil && requestRetrier.ShouldRetryFunc(response) {
```
warber, line 238: nit:
requestRetrier could offer a method to check for retries e.g. like 
```
func (r *RequestRetrier) ShouldRetry(resp *http.Response) bool {
	if r.ShouldRetryFunc != nil {
		return r.ShouldRetryFunc(resp)
	}
	return false
}
```

```suggestion
		requestRetrier.ShouldRetry(resp) {
```

warber (approved):

api/rest/client.go
```diff
@@ -225,10 +234,9 @@ func (c *Client) sendWithRetries(ctx context.Context, req *http.Request, retryCo
 		c.rateLimiter.Update(ctx, response.StatusCode, response.Header)
 	}
 
-	if c.requestRetrier != nil && retryCount < c.requestRetrier.MaxRetries &&
-		c.requestRetrier.ShouldRetryFunc != nil && c.requestRetrier.ShouldRetryFunc(response) {
-		logger.V(1).Info(fmt.Sprintf("Retrying failed request %q (HTTP %s) after %d ms delay... (try %d/%d)", req.URL, response.Status, 100, retryCount+1, c.requestRetrier.MaxRetries), "statusCode", response.StatusCode, "try", retryCount+1, "maxRetries", c.requestRetrier.MaxRetries)
-		time.Sleep(100 * time.Millisecond)
+	if requestRetrier != nil && requestRetrier.ShouldRetry(response, retryCount) {
```
warber, line 237:
```suggestion
	if  requestRetrier.ShouldRetry(response, retryCount) {
```
nit: 
cool, now you could also eliminate the `nil` check by checking for a `nil` receiver in `ShouldRetry()`

```
func (r *RequestRetrier) ShouldRetry(...) bool {
	if r == nil {
		return false
	}
```

warber (approved):

api/rest/client.go
arthurpitman: That might be a step too far.

PR #127, opened 2024-08-27
feat: Simplify overriding retry behavior of REST client

arthurpitman:
This PR simplifies overriding retry behavior to just providing the "should retry function" rather than the whole RequestRetrier structure. This is useful because it makes it easy to change the retry behavior without knowledge of the delay or maximum number of retries.

The PR also adjusts the retry behavior of the clients:
 - essentially turning it off for buckets - this client has its own retry implementation - only retry for `StatusTooManyRequests`
 - for automation client:
   - if doing an "admin access" request, no retrying occurs if `StatusForbidden` is returned,
   - if deleting, no retrying occurs if `StatusNotFound` is returned.
 - for documents client, if deleting, no retrying occurs if `StatusNotFound` is returned

### DIFF
--- a/api/rest/retry.go
+++ b/api/rest/retry.go
@@ -19,25 +19,26 @@ import (
 	"time"
 )
 
+type RetryFunc func(resp *http.Response) bool
+
 // RequestRetrier represents a component for retrying failed HTTP requests.
 type RequestRetrier struct {
 	DelayAfterRetry time.Duration
 	MaxRetries      int
-	ShouldRetryFunc func(resp *http.Response) bool
+	ShouldRetryFunc RetryFunc
 }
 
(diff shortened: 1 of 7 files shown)


api/rest/retry.go
```diff
@@ -19,25 +19,26 @@ import (
 	"time"
 )
 
+type RetryFunc func(resp *http.Response) bool
+
 // RequestRetrier represents a component for retrying failed HTTP requests.
 type RequestRetrier struct {
```
warber, line 25: naming nit: Actually, `RetryOptions` would be a better name.

warber (approved):

warber (approved):

PR #131, opened 2024-09-19
Fix: stop ignore concurrent limit in client factory

jskelin:
This change fixes the issue of ignoring the provided concurrent request limit.

### DIFF
--- a/clients/factory.go
+++ b/clients/factory.go
@@ -128,9 +128,17 @@ func (f factory) WithRetryOptions(retryOptions *rest.RetryOptions) factory {
 	return f
 }
 
-// AccountClient creates and reaturns a new instance of accounts.Client for interacting with the accounts API.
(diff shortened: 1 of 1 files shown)


clients/factory.go
```diff
@@ -226,14 +223,7 @@ func (f factory) createRestClient(u string, httpClient *http.Client) (*rest.Clie
...
-	if f.rateLimiterEnabled {
-		opts = append(opts, rest.WithRateLimiter())
-	}
-
-	if f.retryOptions != nil {
-		opts = append(opts, rest.WithRetryOptions(f.retryOptions))
-	}
+	opts := f.restOptions()
 
 	restClient := rest.NewClient(parsedURL, httpClient, opts...)
```
arthurpitman, line 228: I guess we don't even need `opts`
```suggestion
	restClient := rest.NewClient(parsedURL, httpClient, f.restOptions()...)
```

arthurpitman (approved):

PR #132, opened 2024-09-20
fix: bump the go version to 1.22.7

jskelin:


jskelin:
the version is updated to newer version with PR#133

PR #135, opened 2024-09-24
feat: grail filter segments client

warber:
This PR introduces the http client for grail filter segments to be used by monaco and terraform.

### DIFF
--- a/api/clients/grailfiltersegements/filtersegments.go
+++ b/api/clients/grailfiltersegements/filtersegments.go
@@ -1,0 +1,114 @@
+// @license
+// Copyright 2024 Dynatrace LLC
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package grailfiltersegements
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
+)
+
+const endpointPath = "platform/storage/filter-segments/v1/filter-segments"
+
+type Client struct {
+	client *rest.Client
+}
+
+func NewClient(client *rest.Client) *Client {
+	c := &Client{
+		client: client,
+	}
+	return c
+}
+
+// List gets the list of the existing filter segments. It uses the ":lean" endpoint to get the minimum necessary data set.
+// If the field 'add-fields' in [rest.RequestOptions.QueryParams] is not specified, it will be set to "EXTERNALID".
+func (c Client) List(ctx context.Context, ro rest.RequestOptions) (*http.Response, error) {
+	path := endpointPath + ":lean" // minimal set of information is enough
+
+	// set default behavior to request for ExternalID
+	if !ro.QueryParams.Has("add-fields") {
+		ro.QueryParams = url.Values{"add-fields": []string{"EXTERNALID"}}
+	}
+
+	r, err := c.client.GET(ctx, path, ro)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the list of filter segments: %w", err)
+	}
+
+	return r, nil
+}
+
+// Get filter-segment with UID.
+// If the field 'add-fields' in [rest.RequestOptions.QueryParams] is not specified, it will be set to "INCLUDES", "VARIABLES", "EXTERNALID", "RESOURCECONTEXT".
+func (c Client) Get(ctx context.Context, id string, ro rest.RequestOptions) (*http.Response, error) {
+	path, err := url.JoinPath(endpointPath, id)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create URL: %w", err)
+	}
+
+	// set default behavior to pick request for all information
+	if !ro.QueryParams.Has("add-fields") {
+		ro.QueryParams = url.Values{"add-fields": []string{"INCLUDES", "VARIABLES", "EXTERNALID", "RESOURCECONTEXT"}}
+	}
+
+	r, err := c.client.GET(ctx, path, ro)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get configuration with ID %s: %w", id, err)
+	}
+
+	return r, nil
+}
+
+func (c Client) Create(ctx context.Context, data []byte, ro rest.RequestOptions) (*http.Response, error) {
+	r, err := c.client.POST(ctx, endpointPath, bytes.NewReader(data), ro)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new filter segment: %w", err)
+	}
+	return r, nil
+}
+
+func (c Client) Update(ctx context.Context, id string, data []byte, ro rest.RequestOptions) (*http.Response, error) {
+	path, err := url.JoinPath(endpointPath, id)
+	if err != nil {
+		return nil, fmt.Errorf("failed to join URL: %w", err)
+	}
+
+	r, err := c.client.PUT(ctx, path, bytes.NewReader(data), ro)
+	if err != nil {
+		return nil, fmt.Errorf("failed to update configuration with ID %s: %w", id, err)
+	}
+	return r, nil
+}
+
+func (c Client) Delete(ctx context.Context, id string, ro rest.RequestOptions) (*http.Response, error) {
+	if id == "" {
+		return nil, fmt.Errorf("id must be non-empty")
+	}
+	path, err := url.JoinPath(endpointPath, id)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create URL: %w", err)
+	}
+
+	r, err := c.client.DELETE(ctx, path, ro)
+	if err != nil {
+		return nil, fmt.Errorf("failed to delete configuration with ID %s: %w", id, err)
+	}
+	return r, nil
+}
--- a/api/clients/grailfiltersegements/filtersegments_test.go
+++ b/api/clients/grailfiltersegements/filtersegments_test.go
@@ -28,6 +28,24 @@ import (
 )
 
 func TestList(t *testing.T) {
+	t.Run("check call", func(t *testing.T) {
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			t.Log(r.URL.String())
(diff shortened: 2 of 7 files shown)


clients/grailfiltersegments/filtersegment_test.go
```diff
@@ -0,0 +1,476 @@
...
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/clients/grailfiltersegments"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/testutils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+func Test_List(t *testing.T) {
```
Laubi, line 34: Naming should be `TestList`

clients/grailfiltersegments/filtersegment_test.go
```diff
@@ -0,0 +1,476 @@
...
+	t.Run("base case", func(t *testing.T) {
```
Laubi, line 35: Only one case, please remove it

clients/grailfiltersegments/filtersegment_test.go
```diff
@@ -0,0 +1,476 @@
...
+			Return(&http.Response{
+				StatusCode: http.StatusOK,
+				Body:       io.NopCloser(strings.NewReader(apiResponse)),
+			}, nil)
+
+		fsClient := grailfiltersegments.NewTestClient(mockClient)
+		actual, err := fsClient.List(context.Background())
+
+		require.NoError(t, err)
+		assert.JSONEq(t, expected, string(actual.Data))
```
Laubi, line 69: nit: The mixing of require and assert is not that consistent

clients/grailfiltersegments/filtersegment_test.go
```diff
@@ -0,0 +1,476 @@
...
+	})
+}
+
+func Test_Get(t *testing.T) {
+	t.Run("Get - no ID given", func(t *testing.T) {
```
Laubi, line 74: The naming should just be `No ID`. The rest of the context follows from the test-name. Same for the other cases

```suggestion
	t.Run("No ID", func(t *testing.T) {
```

clients/grailfiltersegments/filtersegment_test.go
```diff
@@ -0,0 +1,476 @@
...
+
+		fsClient := grailfiltersegments.NewTestClient(mockClient)
+		actual, err := fsClient.List(context.Background())
+
+		require.NoError(t, err)
+		assert.JSONEq(t, expected, string(actual.Data))
+	})
+}
+
+func Test_Get(t *testing.T) {
```
Laubi, line 73: The setups in this test are all quite the same, shouldn't we use table tests to make it more easy to view?

clients/grailfiltersegments/filtersegments.go
```diff
@@ -0,0 +1,228 @@
...
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		logr.FromContextOrDiscard(ctx).Error(err, bodyReadErrMsg)
+		return Response{}, api.NewAPIErrorFromResponseAndBody(resp, body)
+	}
+	if !rest.IsSuccess(resp) {
+		return Response{}, api.NewAPIErrorFromResponseAndBody(resp, body)
+	}
+	{
```
Laubi, line 97: Unnecessary section

clients/grailfiltersegments/filtersegments.go
```diff
@@ -0,0 +1,228 @@
...
+		var tmp map[string]any
```
Laubi, line 98: I am sure we find a better name than `tmp` :)

clients/grailfiltersegments/filtersegments.go
```diff
@@ -0,0 +1,228 @@
...
+		if err = json.Unmarshal(body, &tmp); err != nil {
+			return Response{}, api.NewAPIErrorFromResponseAndBody(resp, body)
+		}
+		if body, err = json.Marshal(tmp["filterSegments"]); err != nil {
```
Laubi, line 102: Could you please eloberate what we're doing here exactly? Are we removing everything except the filterSegments? If yes, why? What else is in there?

clients/grailfiltersegments/filtersegments.go
```diff
@@ -0,0 +1,228 @@
...
+	if !rest.IsSuccess(resp) {
+		return Response{}, api.NewAPIErrorFromResponse(resp)
+	}
+	return api.NewResponseFromHTTPResponse(resp), nil
+}
+
+var getRequestOptions = rest.RequestOptions{
+	CustomShouldRetryFunc: rest.RetryIfTooManyRequests,
+	QueryParams: map[string][]string{
+		"add-fields": {"INCLUDES", "VARIABLES", "RESOURCECONTEXT"},
```
Laubi, line 226: These differ from the other fields, why is that? Is it not necessary?

Laubi (changes requested):

Laubi:
To the points mentioned above, please also add documentation to all public methods, with all the assumptions and special cases we're taking

api/clients/grailfiltersegements/filtersegments_test.go
```diff
@@ -0,0 +1,97 @@
+//go:build live
```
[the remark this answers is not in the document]
jskelin: Since it is not wired, I removed this file and will save it locally. When we're going to wire it, I'll add it. This produces more confusion than benefits.

This file is just a tiny wrapper that exposes basic HTTP operations, and it has no logic. For that reason, there is no need to complicate with unit tests

clients/grailfiltersegments/filtersegments.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	return c
+}
+
+// Client can be used to interact with the Automation API
+type Client struct {
+	client client
+}
+
+//go:generate mockgen -source filtersegments.go -package=grailfiltersegments -destination=client_mock.go
```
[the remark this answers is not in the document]
jskelin: It's used. e.g. filtersegment_test.go:57

clients/grailfiltersegments/export_test.go
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package grailfiltersegments
+
+func NewTestClient(client client) *Client {
```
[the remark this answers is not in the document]
jskelin: To reduce complexity and make it simpler for a user, the "proper" API accepts `*github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest.Client`, which is then used to create `grailfiltersegements. Client`. So, with that in place, it isn't a shortcut.

clients/grailfiltersegments/filtersegments.go
jskelin: Moved to separate function. It is also done in description of the function

clients/grailfiltersegments/filtersegments.go
jskelin: By default, not all parts of the configuration are included. If a user doesn't specify what exactly it needs, we assume that it requests everything.

Laubi (changes requested):
Major points that are still required: 
1) 0% test coverage of the filtersegments. I gave an example in my comments.
  a) This is because of the mocks. We should do fakes instead of mocks and use a local server and respond the HTTP results directly. Then, the inner code would be tested.
  b) Because of that, a real bug was not detected that should have been easily detectable, as it's not even in the method doc.

2) Inconsistent error return values. Sometimes you return a plain `err`, sometimes you wrap it. Please consistently wrap it.

3) The inner test names are not consistent in `clients/grailfiltersegments/filtersegment_test.go`

clients/grailfiltersegments/filtersegments.go
jskelin: Moved to the raw client

jskelin (approved):

jskelin (dismissed):

clients/grailfiltersegments/export_test.go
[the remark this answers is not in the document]
jskelin: Can you elaborate on why mocking is a bad practice? Or, to be more precise, what mocking here brings negative

api/clients/grailfiltersegements/filtersegments.go
```diff
@@ -0,0 +1,112 @@
...
+
+	r, err := c.client.GET(ctx, path, ro)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the list of filter segments: %w", err)
+	}
+
+	return r, nil
+}
+
+// Get filter-segment by UID. If query parameter "add-fields" is not specified, it will be set with next values: "INCLUDES", "VARIABLES", "EXTERNALID", "RESOURCECONTEXT".
```
Laubi, line 57:
```suggestion
// // Get filter-segment by its ID.
// If the field 'add-fields' in [rest.RequestOptions.QueryParams] is not specified, it will be set to "INCLUDES", "VARIABLES", "EXTERNALID", "RESOURCECONTEXT".
```

clients/grailfiltersegments/filtersegments.go
```diff
@@ -0,0 +1,214 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/clients/grailfiltersegements"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
+)
+
+const bodyReadErrMsg = "unable to read API response body"
```
Laubi, line 30: This here throws an error as it's unused

Laubi (commented):
Some typos & 2 issues

clients/grailfiltersegments/export_test.go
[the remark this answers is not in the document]
tomazpu: I think this could be a topic for a TechTime, going on a list.

clients/grailfiltersegments/filtersegments.go
```diff
@@ -0,0 +1,212 @@
...
+	return processResponse(resp, normalizeListResponse)
+}
+
+// normalizeListResponse transform received json response to contain just a JSON list of elements
+func normalizeListResponse(source []byte) ([]byte, error) {
+	var transformed map[string]any
+	if err := json.Unmarshal(source, &transformed); err != nil {
+		return source, err
+	}
+	body, err := json.Marshal(transformed["filterSegments"])
```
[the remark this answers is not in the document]
jskelin: In the case of an error, it would return nil, not the original payload. In case of an error, original payload should end up in the APIError

tomazpu (approved):

PR #147, opened 2024-10-24
chore: Update generated account management API

arthurpitman:
This PR updates the generated account management API with the aim of making as few changes to the generated output as possible:

- It updates the generation script `./gen/gen.sh` to use a check-in version of the spec to force tracking what has changed. It is expected that the lasted Open API spec is downloaded and stored as `spec.json` and then formatted and fixed; this script now operates on `./specs/account_management/spec_formatted_fixed.json`. Furthermore we use the option `disallowAdditionalPropertiesIfNotPresent=false`, as [described here](https://openapi-generator.tech/docs/generators/go#config-options).

- Add the most recent versions of the spec, as well as a formatted (and "fixed" version). Adds the generated output.

Further, the PR carries out some steps that hopefully won't be needed in the future:

- It removes some duplicate functions, `reportError(...)` and `newStrictDecoder(...)`

- Patch the implementation of `parameterAddToHeaderOrQuery(...)`. Use the implementation here: https://github.com/OpenAPITools/openapi-generator/blob/654f62ce3c3c32257d8b0e9929fd3fd47f2de15d/samples/openapi3/client/petstore/go/go-petstore/client.go#L164-L251

-  Patch the spec to take care of some inconsistencies: some fields in the API responses are listed as required but they are not included, breaking validation, so these are manually adjusted.


jskelin (approved):

tomazpu (approved):

jskelin (approved):

jskelin:
What's the difference between `spec.json` and `spec_formated.json`?
In case the formatting is the only one, there is no point to save them both.
And, for future proof, I would add the version (or date if the version is not enough)

PR #158, opened 2025-01-03
refactor: Improve segments clients

arthurpitman:
This PR suggests a few minor changes to the segments client found while reviewing other code in the area:
-  makes several error messages more consistent
- fixes a typo in segments package name
- moves checks for empty IDs to the segments API client

### DIFF
--- a/clients/segments/segment_test.go
+++ b/clients/segments/segment_test.go
@@ -76,16 +76,6 @@ func TestList(t *testing.T) {
 }
 
 func TestGet(t *testing.T) {
-	t.Run("call without ID returns an error", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


clients/segments/segment_test.go
```diff
@@ -460,7 +466,13 @@ func TestUpsert(t *testing.T) {
 func TestDelete(t *testing.T) {
 	t.Run("call without ID returns an error", func(t *testing.T) {
```
jskelin, line 467: I would argue that this is testing of API client functionality. So how about moving to `api/clients/segments` ?

jskelin (approved):

PR #162, opened 2025-01-14
feat(documents): Allow arbitrary document types

Laubi:
The client now allows arbitrary types to be set. It's the consumer's responsibility to restrict the types that can be used. There is no change in functionality, as consumers already could use this by casting an arbitrary string to the `DocumentType` type.

### DIFF
--- a/clients/documents/documents.go
+++ b/clients/documents/documents.go
@@ -30,19 +30,23 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
+
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/clients/documents"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
-	"github.com/go-logr/logr"
 )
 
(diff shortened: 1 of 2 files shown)


tomazpu (approved):

clients/documents/documents.go
```diff
@@ -30,19 +30,23 @@ import (
...
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
-	"github.com/go-logr/logr"
 )
 
 const bodyReadErrMsg = "unable to read API response body"
 
-type DocumentType string
+// DocumentType defines the *known* types of documents. It is possible to pass an arbitrary string in consumers
+// to download any kind of document.
+type DocumentType = string
```
jskelin, line 44: I prefer to have a dedicated type over an alias for string.
With dedicated type, a user on the caller side should/ can treat that information as a type and rely on it even at compile time. For example, in Monaco CLI, during the unmarshalling of the config JSON, the consumer should assign the correct value to the appropriate type (not a simple string).

I don't see what pain you resolve with switching to string alias.
@arthurpitman , do you have an opinion on this?

jskelin (commented):
The logic looks fine!

PR #166, opened 2025-01-24
chore(segments): split segments `upsert` into `create` and `update`

tomazpu:
Split segments `upsert` into `create` and `update` to enable usage for Terraform and keep consistency with other implementations of clients.

### DIFF
--- a/clients/segments/segment_test.go
+++ b/clients/segments/segment_test.go
@@ -269,39 +269,7 @@ func TestGetAll(t *testing.T) {
 	})
 }
 
-func TestUpsert(t *testing.T) {
-
-	t.Run("unexpected error while checking for status on server", func(t *testing.T) {
-		payload := `{
-  "uid": "qW5qn449RsG",
-  "name": "dev_environment",
-  "description": "only includes data of the dev environment",
-  "variables": {
-    "type": "query",
-    "value": "fetch logs | limit 1"
-  },
-  "isPublic": false,
-  "owner": "2f321c04-566e-4779-b576-3c033b8cd9e9",
-  "allowedOperations": [
-    "READ",
-    "WRITE",
-    "DELETE"
-  ],
-  "includes": []
-}`
-		ctx := testutils.ContextWithLogger(t)
-		mockClient := segments.NewMockclient(gomock.NewController(t))
-		mockClient.EXPECT().
-			Get(ctx, "uid", gomock.Any()).
-			Return(nil, errors.New("some unexpected error"))
-
-		fsClient := segments.NewTestClient(mockClient)
-		actual, err := fsClient.Upsert(ctx, "uid", []byte(payload))
-
-		require.Error(t, err)
-		require.Empty(t, actual)
-	})
-
+func TestCreate(t *testing.T) {
 	t.Run("successfully created new entity on server", func(t *testing.T) {
 		payload := `{
   "name": "dev_environment",
@@ -342,12 +310,46 @@ func TestUpsert(t *testing.T) {
 			}, nil)
 
 		fsClient := segments.NewTestClient(mockClient)
-		resp, err := fsClient.Upsert(ctx, "", []byte(payload))
+		resp, err := fsClient.Create(ctx, []byte(payload))
 
 		assert.NotEmpty(t, resp)
 		assert.NoError(t, err)
 		assert.Equal(t, apiResponse, string(resp.Data))
 		assert.Equal(t, http.StatusCreated, resp.StatusCode)
+	})
+}
+
+func TestUpdate(t *testing.T) {
+
+	t.Run("unexpected error while checking for status on server", func(t *testing.T) {
+		payload := `{
+  "uid": "qW5qn449RsG",
+  "name": "dev_environment",
+  "description": "only includes data of the dev environment",
+  "variables": {
+    "type": "query",
+    "value": "fetch logs | limit 1"
+  },
+  "isPublic": false,
+  "owner": "2f321c04-566e-4779-b576-3c033b8cd9e9",
+  "allowedOperations": [
+    "READ",
+    "WRITE",
+    "DELETE"
+  ],
+  "includes": []
+}`
+		ctx := testutils.ContextWithLogger(t)
+		mockClient := segments.NewMockclient(gomock.NewController(t))
+		mockClient.EXPECT().
+			Get(ctx, "uid", gomock.Any()).
+			Return(nil, errors.New("some unexpected error"))
+
+		fsClient := segments.NewTestClient(mockClient)
+		actual, err := fsClient.Update(ctx, "uid", []byte(payload))
+
+		require.Error(t, err)
+		require.Empty(t, actual)
 	})
 
 	t.Run("successfully updated existing entity on server", func(t *testing.T) {
@@ -409,7 +411,7 @@ func TestUpsert(t *testing.T) {
 			})
 
 		fsClient := segments.NewTestClient(mockClient)
-		resp, err := fsClient.Upsert(ctx, "uid", []byte(payload))
+		resp, err := fsClient.Update(ctx, "uid", []byte(payload))
 
 		assert.NotEmpty(t, resp)
 		assert.NoError(t, err)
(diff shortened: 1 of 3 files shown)


clients/segments/segments.go
```diff
@@ -113,27 +113,16 @@ func (c Client) GetAll(ctx context.Context) ([]Response, error) {
...
-	var err error
-	if id != "" {
-		existing, err = c.client.Get(ctx, id, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
-		closeBody(existing)
-		if err != nil {
-			return Response{}, fmt.Errorf(errMsgWithId, "upsert", id, err)
-		}
+// Update the matching resource on remote, before the update get is called to ge the version and owner field to be added
+func (c Client) Update(ctx context.Context, id string, data []byte) (Response, error) {
+	if id == "" {
```
jskelin, line 118: The rest of the code is not checking it. It relies on that the core client will do that (e.g., openpipeline, buckets)
Maybe we should discuss it in TecTime.

clients/segments/segments.go
```diff
@@ -148,16 +137,16 @@ func (c Client) Upsert(ctx context.Context, id string, data []byte) (Response, e
...
-		return Response{}, fmt.Errorf(errMsgWithId, "upsert", id, err)
+		return Response{}, fmt.Errorf(errMsgWithId, "update", id, err)
 	}
 	if getResponse.Version == 0 {
 		return Response{}, fmt.Errorf("missing version field in API response")
 	}
 
-	//Adds uid and owner if not set
+	//Adds uid and owner if not set(they are mandatory from the API)
 	data, err = addOwnerAndUIDIfNotSet(data, getResponse.Owner, id)
```
jskelin, line 147: Is it useful to let a user of this method have the ability to set the owner and version manually?

If we expect the user to handle version and owner, it's cheaper to see if it is already set. And only if it isn't go to the server and get that information (it's more important to optimize the network than cpu consumption)

clients/segments/segment_test.go
```diff
@@ -269,39 +269,7 @@ func TestGetAll(t *testing.T) {
...
-			Return(nil, errors.New("some unexpected error"))
-
-		fsClient := segments.NewTestClient(mockClient)
-		actual, err := fsClient.Upsert(ctx, "uid", []byte(payload))
-
-		require.Error(t, err)
-		require.Empty(t, actual)
-	})
-
+func TestCreate(t *testing.T) {
```
jskelin, line 272: Because of editing the given payload, I would like to have a test which tests (and documents) what has the precedence, given or get value for `owner` and `version`.

clients/segments/segment_test.go
```diff
@@ -342,13 +310,47 @@ func TestUpsert(t *testing.T) {
...
+		resp, err := fsClient.Create(ctx, []byte(payload))
 
 		assert.NotEmpty(t, resp)
 		assert.NoError(t, err)
 		assert.Equal(t, apiResponse, string(resp.Data))
 		assert.Equal(t, http.StatusCreated, resp.StatusCode)
 	})
+}
+
+func TestUpdate(t *testing.T) {
```
jskelin, line 322: Because of editing the given payload, I would like to have a test which tests (and documents) what has the precedence, given or get value for `owner` and `version`.

jskelin (approved):
I am for replacing `Upsert`. It would be beneficial to get more approval for that kind of change

clients/segments/segments.go
tomazpu: The handling is for only `owner` and `uid`. So in any case, we need to make a get for the `version`. So no call can be saved there.

clients/segments/segments.go
tomazpu: No need for TechTime, i just didn't see the get checks for empty id. :)

Laubi (approved):

clients/segments/segments.go
```diff
@@ -113,27 +113,18 @@ func (c Client) GetAll(ctx context.Context) ([]Response, error) {
...
-		closeBody(existing)
-		if err != nil {
-			return Response{}, fmt.Errorf(errMsgWithId, "upsert", id, err)
-		}
+// Update puts the content of the segment to the server using http PUT to update segment by id.
+// In the first step GET is called to get mandatory data by the API(owner, uid and version), this is then added to payload.
+func (c Client) Update(ctx context.Context, id string, data []byte) (Response, error) {
+	existing, err := c.client.Get(ctx, id, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+	if existing != nil {
+		defer existing.Body.Close()
```
jskelin, line 121: In case of an error, the bare client returns `(nil, error)` pair, and this will panic.
Either checko for nil, or use `closeBody` which does that

tomazpu: good catch 👍

clients/segments/segments.go
```diff
@@ -113,27 +113,16 @@ func (c Client) GetAll(ctx context.Context) ([]Response, error) {
...
-	if id != "" {
-		existing, err = c.client.Get(ctx, id, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
-		closeBody(existing)
-		if err != nil {
-			return Response{}, fmt.Errorf(errMsgWithId, "upsert", id, err)
-		}
+// Update puts the content of the segment to the server using http PUT to update segment by id.
+// In the first step GET is called to get mandatory data by the API(owner, uid and version), this is then added to payload.
+func (c Client) Update(ctx context.Context, id string, data []byte) (Response, error) {
+	existing, err := c.client.Get(ctx, id, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
```
jskelin, line 119: Reusing the `Get` method from this client would be beneficial.

clients/segments/segments.go
```diff
@@ -148,16 +137,16 @@ func (c Client) Upsert(ctx context.Context, id string, data []byte) (Response, e
...
 	if err != nil {
-		return Response{}, fmt.Errorf(errMsgWithId, "upsert", id, err)
+		return Response{}, fmt.Errorf(errMsgWithId, "update", id, err)
 	}
 	if getResponse.Version == 0 {
 		return Response{}, fmt.Errorf("missing version field in API response")
 	}
 
-	//Adds uid and owner if not set
+	//Adds uid and owner if not set(they are mandatory from the API)
```
jskelin, line 146: Why check the `version` and not the `owner`. Check both or non!

clients/segments/segment_test.go
```diff
@@ -456,24 +458,24 @@ func TestUpsert(t *testing.T) {
...
-			Update(ctx, "uid", gomock.Any(), gomock.AssignableToTypeOf(rest.RequestOptions{})).
+			Update(ctx, "D82a1jdA23a", gomock.Any(), gomock.AssignableToTypeOf(rest.RequestOptions{})).
 			Return(&http.Response{
 				StatusCode: http.StatusOK,
 				Body:       io.NopCloser(strings.NewReader(apiResponse)),
 			}, nil).
 			Do(func(_, _, payload any, ro any) {
 				assertVersionParam(t, ro, "2")
-				assertRequestPayload(payload, t, "uid", "2f321c04-566e-4779-b576-3c033b8cd9e9")
+				assertRequestPayload(payload, t, "D82a1jdA23a", "2f321c04-566e-4779-b576-3c033b8cd9e9")
```
jskelin, line 474: nit: This check is only for `owner` and `UID`. This won't catch the bug where the given payload is overwritten with `Owner` and `UUID`, and the given payload is lost.

clients/segments/segments.go
```diff
@@ -113,27 +113,16 @@ func (c Client) GetAll(ctx context.Context) ([]Response, error) {
...
-	var err error
-	if id != "" {
-		existing, err = c.client.Get(ctx, id, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
-		closeBody(existing)
-		if err != nil {
-			return Response{}, fmt.Errorf(errMsgWithId, "upsert", id, err)
-		}
+// Update puts the content of the segment to the server using http PUT to update segment by id.
+// In the first step GET is called to get mandatory data by the API(owner, uid and version), this is then added to payload.
+func (c Client) Update(ctx context.Context, id string, data []byte) (Response, error) {
```
jskelin, line 118: Should we return the updated segment config?

clients/segments/segments.go
```diff
@@ -148,16 +137,16 @@ func (c Client) Upsert(ctx context.Context, id string, data []byte) (Response, e
 
 	err = json.Unmarshal(existingResourceBody, &getResponse)
 	if err != nil {
-		return Response{}, fmt.Errorf(errMsgWithId, "upsert", id, err)
+		return Response{}, fmt.Errorf(errMsgWithId, "update", id, err)
 	}
 	if getResponse.Version == 0 {
```
jskelin, line 142: Why check the version and not the owner. Check both or non!

clients/segments/segments.go
```diff
@@ -113,27 +113,16 @@ func (c Client) GetAll(ctx context.Context) ([]Response, error) {
...
-		existing, err = c.client.Get(ctx, id, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
-		closeBody(existing)
-		if err != nil {
-			return Response{}, fmt.Errorf(errMsgWithId, "upsert", id, err)
-		}
+// Update puts the content of the segment to the server using http PUT to update segment by id.
+// In the first step GET is called to get mandatory data by the API(owner, uid and version), this is then added to payload.
+func (c Client) Update(ctx context.Context, id string, data []byte) (Response, error) {
+	existing, err := c.client.Get(ctx, id, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+	closeBody(existing)
```
jskelin, line 120: This won't work. Either defer it or move after you consume the reader of the body

In the test, we have `io.NopCloser(strings.NewReader(apiExistingResource))`.

jskelin (approved):

PR #168, opened 2025-01-30
feat(slo-v2): add client

jskelin:
Bare and smart clients are extended with standard methods to support the slo-v2 endpoint. 

Smart client contains next methods:
*  `List(ctx context.Context) (api.PagedListResponse, error)`
*  `Get(ctx context.Context, id string) (api.Response, error)`
*  `Create(ctx context.Context, body json.RawMessage) (api.Response, error)`
*  `Update(ctx context.Context, id string, body json.RawMessage) (api.Response, error)`
*  `Delete(ctx context.Context, id string) (api.Response, error)`


clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+
+// List gets a complete set of complete configuration for all available SLOs
+func (c Client) List(ctx context.Context) (api.PagedListResponse, error) {
+	var retVal api.PagedListResponse
+
+	for haveNextPage, nextPageKey := true, ""; haveNextPage; {
+		response, err := c.client.List(ctx, rest.RequestOptions{
+			CustomShouldRetryFunc: rest.RetryIfTooManyRequests,
+			QueryParams:           url.Values{"page-key": {nextPageKey}}})
+		defer closeBody(response)
```
Laubi, line 62: This (inner loop) should be extracted in a function, as this is a minor resource leak (body stays open until all resources have been downloaded)

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+	resp, err := c.client.Update(ctx, id, version, body, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+	defer closeBody(resp)
+	if err != nil {
+		return api.Response{}, fmt.Errorf(errMsgWithId, "update", id, err)
+	}
+	return processResponse(resp, nil)
+}
+
+func getVersion(resp api.Response) (string, error) {
+	var getStr struct {
```
Laubi, line 119: What does `getStr` mean?

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+		Version string `json:"version"`
+	}
+
+	err := json.Unmarshal(resp.Data, &getStr)
+	if err != nil {
+		return "", err
```
Laubi, line 125: should probablty add more information here?

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+		return api.Response{}, fmt.Errorf(errMsgWithId, "delete", id, err)
+	}
+
+	if !rest.IsSuccess(resp) {
+		return api.Response{}, api.NewAPIErrorFromResponse(resp)
+	}
+	return api.NewResponseFromHTTPResponseAndBody(resp, nil), nil
+}
+
+func closeBody(httpResponse *http.Response) {
```
Laubi, line 145: We have this call 1:1 in other files, we should abstract this somehow

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+	if httpResponse != nil && httpResponse.Body != nil {
+		_ = httpResponse.Body.Close()
+	}
+}
+
+func processResponse(httpResponse *http.Response, transform func([]byte) ([]byte, error)) (api.Response, error) {
```
Laubi, line 151: transform is always nil

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+	if transform != nil {
+		if body, err = transform(body); err != nil {
+			return api.Response{}, api.NewAPIErrorFromResponseAndBody(httpResponse, body)
+		}
+	}
+
+	return api.NewResponseFromHTTPResponseAndBody(httpResponse, body), nil
+}
+
+type nextPage = string
```
Laubi, line 174: I don't think having this is necessary. Where it's used should probably used named returns to clarify the meaning. Having an extra type here has IMO no benefits.

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+
+func processListResponse(httpResponse *http.Response, transform func(json.RawMessage) (nextPage string, data [][]byte, e error)) (nextPage, api.ListResponse, error) {
```
Laubi, line 176: You call this method with a fixed method, and it's only used once. Why pass it?

Laubi (changes requested):
I'd like to discuss some things that I quickly found. I for sure need a second round eventually: 

1) The 'bare' client makes no sense to me.
2) We should discuss the interface of the client and agree on one usage.  The segments client has `GetAll` and `List`, but `List` is not even used in Monaco. Somehow we have a disconnect.
2) I really thought we decided to no longer use mocks for new APIs and eventually get rid of the old ones.

I did not yet reviewed the test in detail, but please use proper English in the tests, and do not only test the sunny test cases.

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+	List(ctx context.Context, ro rest.RequestOptions) (*http.Response, error)
+	Get(ctx context.Context, id string, ro rest.RequestOptions) (*http.Response, error)
+	Create(ctx context.Context, data []byte, ro rest.RequestOptions) (*http.Response, error)
+	Update(ctx context.Context, id string, optimisticLockingVersion string, data []byte, ro rest.RequestOptions) (*http.Response, error)
+	Delete(ctx context.Context, id string, ro rest.RequestOptions) (*http.Response, error)
+}
+
+var _ client = (*slo.Client)(nil)
+
+const errMsgWithId = "failed to %s slo resource with id %s: %w"
```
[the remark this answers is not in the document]
tomazpu: I agree, but I would do it in a follow-up to other APIs too.

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+	resp, err := c.client.Create(ctx, body, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+	defer closeBody(resp)
+	if err != nil {
+		return api.Response{}, err // error message from the client is descriptive enough
+	}
+	return processResponse(resp, nil)
+}
+
+func (c Client) Update(ctx context.Context, id string, body json.RawMessage) (api.Response, error) {
+	getResp, err := c.Get(ctx, id)
```
tomazpu, line 100: `getResp` body is not closed.

clients/slo/slo.go
tomazpu: Nit, i would also call it something more meaningful.

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+		return "", err
+	}
+
+	return getStr.Version, nil
+}
+
+// Delete removes configuration for SLO with given ID from a server.
+func (c Client) Delete(ctx context.Context, id string) (api.Response, error) {
+	resp, err := c.client.Delete(ctx, id, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+	closeBody(resp)
```
tomazpu, line 134: defer keyword missing

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+
+	if !rest.IsSuccess(resp) {
+		return api.Response{}, api.NewAPIErrorFromResponse(resp)
+	}
+	return api.NewResponseFromHTTPResponseAndBody(resp, nil), nil
+}
+
+func closeBody(httpResponse *http.Response) {
+	if httpResponse != nil && httpResponse.Body != nil {
+		_ = httpResponse.Body.Close()
```
tomazpu, line 147: I wonder if we should handle the error. Just an opinion, what do you think?

api/clients/slo/slo.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (c Client) Create(ctx context.Context, data []byte, ro rest.RequestOptions) (*http.Response, error) {
+	r, err := c.restClient.POST(ctx, endpointPath, bytes.NewReader(data), ro)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new SLO: %w", err)
+	}
+	return r, nil
+}
+
+func (c Client) Update(ctx context.Context, id string, optimisticLockingVersion string, data []byte, ro rest.RequestOptions) (*http.Response, error) {
```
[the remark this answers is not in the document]
jskelin: In the meeting, we decided that the core library will no longer have bare clients. This is moved as an internal part of the smart client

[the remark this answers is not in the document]
jskelin: The bare client is removed. The suggestion is outdated.

api/clients/slo/slo.go
```diff
@@ -0,0 +1,115 @@
...
+	if id == "" {
+		return nil, fmt.Errorf("id must be non-empty")
+	}
+	if optimisticLockingVersion == "" {
+		return nil, fmt.Errorf("optimisticLockingVersion must be non-empty")
```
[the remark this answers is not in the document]
jskelin: The bare client is removed. The suggestion is outdated.

api/clients/slo/slo.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	path, err := url.JoinPath(endpointPath, id)
+	if err != nil {
+		return nil, fmt.Errorf("failed to join URL: %w", err)
```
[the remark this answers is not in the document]
jskelin: The bare client is removed. The suggestion is outdated.

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+		client: slo.NewClient(client),
+	}
+	return c
+}
+
+type Client struct {
+	client client
+}
+
+//go:generate mockgen -source slo.go -package=slo -destination=client_mock.go
```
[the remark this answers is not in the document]
jskelin: removed

api/clients/slo/slo_test.go
```diff
@@ -0,0 +1,165 @@
...
+		ctx := testutils.ContextWithLogger(t)
+		resp, err := client.Update(ctx, "ID", "", []byte{}, rest.RequestOptions{})
+		assert.Zero(t, resp)
+		assert.ErrorContains(t, err, "optimisticLockingVersion")
+	})
+
+}
+
+func TestDelete(t *testing.T) {
+	t.Run("check call", func(t *testing.T) {
```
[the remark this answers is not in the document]
jskelin: The bare client is removed. The suggestion is outdated.

clients/slo/slo.go
jskelin: done

clients/slo/slo.go
jskelin: I renamed it to body, but I'm not happy with this name either. Can you think of a better one?

clients/slo/slo.go
jskelin: This is part of a separate PR: https://github.com/Dynatrace/dynatrace-configuration-as-code-core/pull/172

clients/slo/slo.go
jskelin: This is part of a separate PR: https://github.com/Dynatrace/dynatrace-configuration-as-code-core/pull/172

clients/slo/slo.go
jskelin: removed

clients/slo/slo.go
jskelin: changed

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+// Delete removes configuration for SLO with given ID from a server.
+func (c Client) Delete(ctx context.Context, id string) (api.Response, error) {
+	resp, err := c.client.Delete(ctx, id, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+	closeBody(resp)
+	if err != nil {
+		return api.Response{}, fmt.Errorf(errMsgWithId, "delete", id, err)
+	}
+
+	if !rest.IsSuccess(resp) {
+		return api.Response{}, api.NewAPIErrorFromResponse(resp)
```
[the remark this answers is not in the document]
jskelin: changed

api/clients/doc.go
```diff
@@ -21,8 +21,8 @@ In general, the CRUD clients make a single API call per operation, but in some c
 
 Clients implement the following methods:
 
-  - Get
   - List
+  - Get
```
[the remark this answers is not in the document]
jskelin: How about to move it to the last one?
I would like that CRUR are together. List and GetAll are different in the sense that they return multiple values. And are  unshaped yet

clients/slo/slo.go
jskelin: Here it's not a `http.Response`. This is a call to the Get method of this client, and `http.Response` is handled there.

clients/slo/slo.go
jskelin: fixed

clients/slo/slo.go
jskelin: I wouldn't. To return an error from a deferred function, we need to use named return values. And then the fun part begins. Just returning it will overwrite the result from the function.
On the other hand, I don't see what we could do in case of an error. The current behavior of just gluing an error on the end and propagating it up is also not beneficial.

clients/slo/export_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+package slo
+
+import "encoding/json"
+
+func NewTestClient(client client) *Client {
+	return &Client{client: client}
+}
+
+func UnmarshallFromListResponse(in json.RawMessage) (string, [][]byte, error) {
```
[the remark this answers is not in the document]
jskelin: removed

clients/slo/slo.go
jskelin: It is a helper function, and the caller adds more information to give the context. And the probability for this error is small.

jskelin: I wouldn't go with more.

clients/slo/slo.go
```diff
@@ -0,0 +1,212 @@
...
+	resp, err := c.restClient.GET(ctx, path, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+	if err != nil {
+		return api.Response{}, fmt.Errorf(errMsgWithId, "get", id, err)
+	}
+	defer resp.Body.Close()
+
+	return api.ProcessResponse(resp)
+}
+
+func (c *Client) Create(ctx context.Context, body json.RawMessage) (api.Response, error) {
```
tomazpu, line 95: Changing `json.RawMessage` from `data []byte` will be a problem in the monaco refactoring. So please leave it as in the other clients `data []byte`.

clients/slo/slo.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+type Client struct {
+	restClient *rest.Client
+}
+
+func (c *Client) List(ctx context.Context) (api.PagedListResponse, error) {
+	var retVal api.PagedListResponse
+
+	listPage := func(ctx context.Context, c *Client, pageKey string) (string, api.ListResponse, error) {
```
Laubi, line 47: Why not move this into its own function?

clients/slo/slo.go
```diff
@@ -0,0 +1,212 @@
...
+		haveNextPage = nextPageKey != ""
+	}
+
+	return retVal, nil
+
+}
+
+func (c *Client) Get(ctx context.Context, id string) (api.Response, error) {
+	if id == "" {
+		return api.Response{}, fmt.Errorf(errMsgWithId, "get", id, errors.New("argument \"id\" is empty"))
```
Laubi, line 78:
```suggestion
		return api.Response{}, fmt.Errorf(errMsgWithId, "get", id, errors.New(`argument "id" is empty`))
```

clients/slo/slo.go
```diff
@@ -0,0 +1,212 @@
...
+
+	err := json.Unmarshal(resp.Data, &body)
+	if err != nil {
+		return "", err
+	}
+
+	return body.Version, nil
+}
+
+func processListResponse(httpResponse *http.Response) (string, api.ListResponse, error) {
```
Laubi, line 185: For 3 or more return values, please add names. I have no idea what the first argument here is without looking at the code

clients/slo/slo.go
```diff
@@ -0,0 +1,212 @@
...
+		return "", api.ListResponse{}, api.NewAPIErrorFromResponseAndBody(httpResponse, body)
+	}
+
+	var s struct {
+		NextPage string            `json:"nextPageKey"`
+		Data     []json.RawMessage `json:"slos"`
+	}
+
+	if err := json.Unmarshal(body, &s); err != nil {
+		return "", api.ListResponse{}, api.NewAPIErrorFromResponseAndBody(httpResponse, body)
```
Laubi, line 203: Error is completely lost why the unmarshalling failed

clients/slo/slo.go
```diff
@@ -0,0 +1,212 @@
...
+		return api.Response{}, fmt.Errorf(errMsg, "create", err)
+	}
+	defer resp.Body.Close()
+
+	return api.ProcessResponse(resp)
+}
+
+func (c *Client) Update(ctx context.Context, id string, body json.RawMessage) (api.Response, error) {
+	if id == "" {
+		return api.Response{}, fmt.Errorf(errMsgWithId, "update", id, errors.New("argument \"id\" is empty"))
```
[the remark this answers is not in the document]
jskelin: That's true. To me it's more valuable to follow the "Convention over Configuration" style. In the case of checking arguments, this should always be the first step, and don't rely on other functions.

clients/slo/slo.go
jskelin: This is C/P. The same situation is on ln 189-190. I would leave it like this: Let's discuss what the desired behavior is and change it for all clients.

clients/slo/slo.go
jskelin: done

clients/slo/slo.go
```diff
@@ -0,0 +1,212 @@
...
+const errMsgWithId = "failed to %s slo resource with id %s: %w"
+
+func getOptimisticLockingVersion(resp api.Response) (string, error) {
+	var body struct {
+		Version string `json:"version"`
+	}
+
+	err := json.Unmarshal(resp.Data, &body)
+	if err != nil {
+		return "", err
```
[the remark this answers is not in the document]
jskelin: added

clients/slo/slo.go
```diff
@@ -0,0 +1,212 @@
...
+	resp, err := c.restClient.POST(ctx, endpointPath, bytes.NewReader(body), rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+	if err != nil {
+		return api.Response{}, fmt.Errorf(errMsg, "create", err)
+	}
+	defer resp.Body.Close()
+
+	return api.ProcessResponse(resp)
+}
+
+func (c *Client) Update(ctx context.Context, id string, body json.RawMessage) (api.Response, error) {
```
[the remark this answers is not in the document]
jskelin: changed

clients/slo/slo.go
```diff
@@ -0,0 +1,212 @@
...
+			QueryParams:           url.Values{"page-key": {pageKey}}})
+		if err != nil {
+			return "", api.ListResponse{}, fmt.Errorf(errMsg, "list", err)
+		}
+		defer resp.Body.Close()
+
+		return processListResponse(resp)
+	}
+
+	for haveNextPage, nextPageKey := true, ""; haveNextPage; {
```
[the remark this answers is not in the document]
jskelin: changed

clients/slo/slo.go
jskelin: changed

clients/slo/slo.go
```diff
@@ -0,0 +1,213 @@
...
+		haveNextPage = nextPageKey != ""
+	}
+
+	return retVal, nil
+}
+
+// Get gets a complete configuration of SLO with an ID
+func (c Client) Get(ctx context.Context, id string) (api.Response, error) {
+	resp, err := c.client.Get(ctx, id, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+	defer closeBody(resp)
```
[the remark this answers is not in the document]
jskelin: moved

api/clients/slo/slo.go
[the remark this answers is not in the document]
jskelin: The bare client is removed. The suggestion is outdated.

clients/slo/slo.go
jskelin: removed

clients/slo/slo.go
jskelin: added

clients/slo/slo.go
jskelin: It's about style.
The whole function fits to a screen (less than 40 lines of code). The `listPage` is only used inside the `List` method, and they're tightly coupled.

clients/slo/slo.go
```diff
@@ -0,0 +1,205 @@
...
+		restClient: client,
+	}
+	return c
+}
+
+type Client struct {
+	restClient *rest.Client
+}
+
+func (c *Client) List(ctx context.Context) (api.PagedListResponse, error) {
```
tomazpu, line 44: Why not just return []ListResponse as PagedListResponse it's an alias anyway.

clients/slo/slo.go
tomazpu: I would agree with david it may be that the lines of code are not to many but, do i need to know the details here? Also another point is it provides a better naming possibility.

clients/slo/slo.go
```diff
@@ -0,0 +1,205 @@
...
+	for hasNextPage, nextPageKey := true, ""; hasNextPage; {
+		var listResponse api.ListResponse
+		var err error
+
+		nextPageKey, listResponse, err = listPage(ctx, c, nextPageKey)
+		if err != nil {
+			return nil, fmt.Errorf(errMsg, "list", err)
+		}
+
+		retVal = append(retVal, listResponse)
```
tomazpu, line 67: the APi returns a totalCount of elements. If we use this in the declaration of the api.PagedListResponse, we can make the append more efficient. About 3x faster if you declare it as ` list := make(api.PagedListResponse, totalCount)`

api/response.go
```diff
@@ -98,6 +98,18 @@ func (p PagedListResponse) All() [][]byte {
 	return ret
 }
 
+func NewListResponse(resp *http.Response, body [][]byte) ListResponse {
```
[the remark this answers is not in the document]
tomazpu: As you introduced this change and it's only used by SLO, I would like to resolve this with the PR.

clients/slo/slo_test.go
```diff
@@ -0,0 +1,465 @@
...
+      ],
+      "tags": [],
+      "id": "slo-id-2",
+      "version": "ver1"
+    }
+  ]
+}`
+
+		server := httptest.NewServer(http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+			require.Equal(t, http.MethodGet, request.Method)
```
tomazpu, line 86: nit: the boilerplate code of the test is the same everywhere, why not inject the behavior in and prevent the repetition?

clients/slo/slo.go
jskelin: resolved in call

clients/slo/slo.go
jskelin: `totalCount` is the number of entries that exist on the server. 
retVal is a set of pages! Those 2 numbers are not equal.

clients/slo/slo.go
jskelin: moved

api/response.go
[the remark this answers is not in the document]
jskelin: As this function is used only once, and bringing more logic into it would bring unnecessary data coping, it is removed.

clients/slo/slo.go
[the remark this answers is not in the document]
jskelin: discused

Kirdock:
Just asking silly questions: Did we decide to now use the factory anymore or is the creation of the SLO client missing in there?

jskelin:
> Just asking silly questions: Did we decide to now use the factory anymore or is the creation of the SLO client missing in there?

Great observation. The factory is added

clients/factory.go
```diff
@@ -183,6 +184,15 @@ func (f factory) SegmentsClient() (*segments.Client, error) {
 	return segments.NewClient(restClient), nil
 }
 
+// SLOClient creates and returns a new instance of slo.Client for interacting with the SLO API.
+func (f factory) SLOClient() (*slo.Client, error) {
```
Laubi, line 188: Follow up: Add to the existing tests in `factory_test.go`

clients/slo/slo.go
```diff
@@ -0,0 +1,220 @@
...
+	if id == "" {
+		return api.Response{}, fmt.Errorf(errMsgWithId, "get", id, errors.New(`argument "id" is empty`))
+	}
+
+	path, err := url.JoinPath(endpointPath, id)
+	if err != nil {
+		return api.Response{}, fmt.Errorf(errMsgWithId, "get", id, err)
+	}
+
+	resp, err := c.restClient.GET(ctx, path, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
```
Laubi, line 88: Follow up: Please add tests if the network breaks, so when GET, POST, etc return an error. 
Imo these are important error cases to test.

Laubi (approved):

clients/factory.go
jskelin: https://github.com/Dynatrace/dynatrace-configuration-as-code-core/pull/173

PR #174, opened 2025-02-18
fix(slo): add slo client API clients in readme.md

tomazpu:
Update readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ The library provides different kinds of clients to interact with Dynatrace in tw
   | Documents           | ✅           |
   | OpenPipeline        | ✅           |
   | Segments            | ✅           |
+  | SLO's               | ✅           |
 
 ### Usage
 


arthurpitman (approved):

README.md
```diff
@@ -27,6 +27,7 @@ The library provides different kinds of clients to interact with Dynatrace in tw
   | Documents           | ✅           |
   | OpenPipeline        | ✅           |
   | Segments            | ✅           |
+  | SLO's               | ✅           |
```
jskelin, line 30: it should be SLOv2

jskelin (changes requested):

README.md
Laubi: I think SLO is fine. From public side, there is no 'SLO v2', it's just classic and platform. We use the naming in Monaco (client) to distinguish it.
We're here at the agnostic core library, and there is no 'slo-v1' client. That's implicit in the Classic config Client.

We could link maybe to the [documentation here](https://docs.dynatrace.com/docs/discover-dynatrace/references/dynatrace-api/environment-api/service-level-objectives) to clarify what API the clients target. If we do that, we should do it for the other APIs as well though. 

IF you want to use a more specific naming, I would rather go for Platform SLOs, rather than v2, but even this is, IMO, not necessary.

PR #176, opened 2025-02-18
chore: Update Go to version 1.24

Kirdock:
Updates go (lib and pipeline) to `1.24`

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -33,13 +33,13 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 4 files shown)


jskelin:
Could you make another push and introduce `go-version-file` (https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file)

Kirdock:
> Could you make another push and introduce `go-version-file` (https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file)

Done. I also had to move the checkout step before setting the go version, to actually get the `go.mod` file

jskelin (approved):

PR #185, opened 2025-03-11
chore: Update Account Management client

arthurpitman:
This PR updates the Account Management OpenAPI spec and uses the openapi-generator to regenerate the client.


gen/specs/account_management/spec_formatted_fixed.json
```diff
@@ -4629,127 +4925,949 @@
...
+            }
+        },
+        "/v1/subscriptions/{subscription-uuid}/cost-allocation": {
+            "get": {
+                "operationId": "InsightsController_getCostAllocationBreakdownPage",
+                "summary": "Get usage data by cost allocation field.",
+                "parameters": [
+                    {
+                        "name": "subscription-uuid",
+                        "required": true,
```
Kirdock, line 5709: Compared to `spec_formatted.json` this is `true` instead of `false`.
Any reason for this change?

arthurpitman: I don't think so, so I will revert it

Kirdock (approved):

PR #188, opened 2025-03-19
feat: introduce not-found error check

Kirdock:
Adds a utility function to check if an error is an api error with status code 404

### DIFF
--- a/api/response_test.go
+++ b/api/response_test.go
@@ -422,6 +422,40 @@ func TestAsResponseOrError(t *testing.T) {
 	})
 }
 
+func TestIsNotFoundError(t *testing.T) {
+	t.Run("404 error returns true", func(t *testing.T) {
+		err := api.APIError{StatusCode: http.StatusNotFound}
+		got := api.IsNotFoundError(err)
+		assert.Equal(t, true, got)
+	})
+
+	t.Run("Different error returns true", func(t *testing.T) {
+		err := api.APIError{StatusCode: 400}
+		got := api.IsNotFoundError(err)
+		assert.Equal(t, false, got)
+	})
+
+	t.Run("Not 404 api error returns false", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


api/response_test.go
```diff
@@ -422,6 +422,40 @@ func TestAsResponseOrError(t *testing.T) {
...
 }
 
+func TestIsNotFoundError(t *testing.T) {
+	t.Run("404 error returns true", func(t *testing.T) {
+		err := api.APIError{StatusCode: http.StatusNotFound}
+		got := api.IsNotFoundError(err)
+		assert.Equal(t, true, got)
+	})
+
+	t.Run("Different error returns true", func(t *testing.T) {
```
arthurpitman, line 432:
```suggestion
	t.Run("Different error returns false", func(t *testing.T) {
```

api/response_test.go
```diff
@@ -422,6 +422,40 @@ func TestAsResponseOrError(t *testing.T) {
...
+		got := api.IsNotFoundError(err)
+		assert.Equal(t, false, got)
+	})
+
+	t.Run("Not api error returns false", func(t *testing.T) {
+		err := customErr{StatusCode: http.StatusNotFound}
+		got := api.IsNotFoundError(err)
+		assert.Equal(t, false, got)
+	})
+}
```
arthurpitman, line 449: Probably self-explanatory, but this also works for wrapped errors:
``` 
t.Run("Wrapped api error with status not found returns true", func(t *testing.T) {
    err := fmt.Errorf("wrapping api error: %w", api.APIError{StatusCode: http.StatusNotFound})
    got := api.IsNotFoundError(err)
assert.Equal(t, true, got)
})
```

tomazpu (approved):
Nothing else then the points @arthurpitman found, so i approve it.

tomazpu (approved):

arthurpitman (approved):

PR #191, opened 2025-03-20
chore: ignore 403 on retry

Kirdock:
Ignore certain status codes that should not be retried (e.g., 403), because they will always fail

### DIFF
--- a/api/rest/client.go
+++ b/api/rest/client.go
@@ -19,12 +19,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/go-logr/logr"
-	"github.com/google/uuid"
 	"io"
 	"net/http"
 	"net/url"
 	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/google/uuid"
 )
 
 // RequestOptions are additional options that should be applied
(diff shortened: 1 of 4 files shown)


api/rest/client.go
```diff
@@ -233,7 +234,7 @@ func (c *Client) sendWithRetries(ctx context.Context, req *http.Request, retryCo
 		c.rateLimiter.Update(ctx, response.StatusCode, response.Header)
 	}
 
-	if c.retryOptions != nil && retryCount < c.retryOptions.MaxRetries {
+	if ShouldRetry(response.StatusCode) && c.retryOptions != nil && retryCount < c.retryOptions.MaxRetries {
```
tomazpu, line 237: Nit: To improve readability, I personally would move this c.retryOptions != nil && retryCount < c.retryOptions.MaxRetrie into the `ShouldRetry` method.

Kirdock: My plan was to make it re-usable also for Monaco

api/rest/retry_test.go
```diff
@@ -0,0 +1,56 @@
...
+		},
+	}
+
+	for _, tt := range testcases {
+		t.Run(fmt.Sprintf("Should be %v if status is %d", tt.want, tt.statusCode), func(t *testing.T) {
+			got := rest.ShouldRetry(tt.statusCode)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
```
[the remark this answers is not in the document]
Kirdock: Removed table test and introduced separate tests

tomazpu (approved):

arthurpitman (approved):

PR #212, opened 2025-05-05
feat(settings): Adding settings object permission client

d0weinberger:

The main goal of this commit is to add the client code for the settings object permission handling API.

This code can then be used by both, the dynatrace-configuration-as-code project and the Terraform provider.

Yes, a new client for managing settings object permissions is provided.

### DIFF
--- a/clients/settings/permissions/permissions.go
+++ b/clients/settings/permissions/permissions.go
@@ -1,0 +1,163 @@
+// @license
+// Copyright 2025 Dynatrace LLC
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package permissions
+
+import (
+	"bytes"
+	"context"
+	"net/url"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
+)
+
+const endpointConfigPath = "platform/classic/environment-api/v2/settings/objects"
+const permissionResourcePath = "permissions"
+const allUsersAccessorType = "all-users"
+
+type Client struct {
+	client *rest.Client
+}
+
+func NewClient(client *rest.Client) *Client {
+	return &Client{client: client}
+}
+
+func (c *Client) GetAllAccessors(ctx context.Context, objectID string) (api.Response, error) {
+	return c.get(ctx, objectID, "", "")
+}
+
+func (c *Client) GetAllUsersAccessor(ctx context.Context, objectID string) (api.Response, error) {
+	return c.get(ctx, objectID, allUsersAccessorType, "")
+}
+
+func (c *Client) GetAccessor(ctx context.Context, objectID string, accessorType string, accessorID string) (api.Response, error) {
+	if accessorType == "" {
+		return api.Response{}, ErrorPermissionGet{ErrorMissingAccessorType}
+	}
+
+	if accessorID == "" {
+		return api.Response{}, ErrorPermissionGet{ErrorMissingAccessorID}
+	}
+
+	return c.get(ctx, objectID, accessorType, accessorID)
+}
+
+func (c *Client) get(ctx context.Context, objectID string, accessorType string, accessorID string) (api.Response, error) {
+	if objectID == "" {
+		return api.Response{}, ErrorPermissionGet{ErrorMissingObjectID}
+	}
+
+	path, err := url.JoinPath(endpointConfigPath, objectID, permissionResourcePath, accessorType, accessorID)
+	if err != nil {
+		return api.Response{}, ErrorPermissionGet{err}
+	}
+
+	resp, err := c.client.GET(ctx, path, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+
+	if err != nil {
+		return api.Response{}, ErrorPermissionGet{err}
+	}
+
+	return api.NewResponseFromHTTPResponse(resp)
+}
+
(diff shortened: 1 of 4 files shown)


d0weinberger:
The failing checks are due to duplicated code - Create and Update look nearly identical. I'm not convinced, though, that we benefit much from trying to unify these functions.

clients/settings/errors.go
Kirdock, line 1: Idea: Should we structure it more generically? For example, adding a context to the error structs ("permission"/"settings", etc.). Then they can also be easily reused without having a massive amount of custom errors

clients/settings/permissions.go
```diff
@@ -0,0 +1,341 @@
...
+	if err != nil {
+		return PermissionObjects{}, err
+	}
+	httpResp, err := d.client.GET(
+		ctx,
+		path,
+		rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+
+	if err != nil {
+		return PermissionObjects{}, err
```
Kirdock, line 119: Should we also add some context here? Custom errors return an error, indicating that it is related to permissions, and the non-custom ones don't return any context.
At the moment, all of the functions are a bit inconsistent regarding this

clients/settings/permissions.go
```diff
@@ -0,0 +1,341 @@
...
+
+	_, err = api.NewResponseFromHTTPResponse(httpResp)
+	if err != nil {
+		return ErrorPermissionUpdate{err}
+	}
+
+	return err
+}
+
+// Delete deletes the permissions of the provided accessor for the settings object identified by the provided objectID.
```
Kirdock, line 285: Maybe also mention that 404 errors are ignored

clients/settings/errors.go
```diff
@@ -0,0 +1,73 @@
...
+package settings
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	ErrorMissingObjectId   = errors.New("objectId cannot be empty")
+	ErrorMissingAccessorId = errors.New("accessorId cannot be empty")
```
arthurpitman, line 24: Do we have an opinion about using sentinel errors vs custom error types?

clients/settings/permissions.go
arthurpitman: True, this is inconsistent. But I imagine the caller would wrap the error anyway so that the context would come there. Therefore, I would prefer to remove the superfluous context in the other errors. I like errors to be as short as possible and have the context as far out as possible.

clients/settings/permissions.go
arthurpitman: Is there a reason we ignore 404s?

clients/settings/permissions_test.go
```diff
@@ -0,0 +1,952 @@
...
+			server := testutils.NewHTTPTestServer(t, []testutils.ResponseDef{responseDef})
+			defer server.Close()
+
+			dcl := settings.NewClient(rest.NewClient(server.URL(), server.Client()))
+			res, err := dcl.GetAll(t.Context(), "12345")
+			assert.NoError(t, err)
+			assert.Equal(t, expectedResponse, res)
+		})
+	})
+	t.Run("error cases", func(t *testing.T) {
```
arthurpitman, line 104: Should we add a case for invalid JSON here and for the other methods?

d0weinberger:
Please excuse the messy commit history here. Based on our discussion earlier, I changed the client such that it only returns api.Response. It is now very similar to the slo client.
I **will fix** the commit history but wanted to get your feedback first before removing the "old" solution from the history.

clients/settings/permissions.go
d0weinberger: All this will be moved away from the core lib.

clients/settings/permissions_test.go
d0weinberger: All this will be moved away from the core lib. Dealing with invalid JSON must be done and tested in monaco repository

clients/settings/permissions.go
d0weinberger: I now always wrap errors in either `ErrorPermissionGet`, `ErrorPermissionCreate`, `ErrorPermissionUpdate`, or `ErrorPermissionsDelete`.

clients/settings/permissions/permissions.go
```diff
@@ -0,0 +1,163 @@
...
+	"context"
+	"net/url"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
+)
+
+const endpointConfigPath = "platform/classic/environment-api/v2/settings/objects"
+const permissionResourcePath = "permissions"
+const allUsersAccessorType = "all-users"
```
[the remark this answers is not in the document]
d0weinberger: Not sure where to use them though.
Maybe for validating the `accessorType` parameter in `GetAccessor`,`UpdateAccessor`, and `DeleteAccessor`.

clients/settings/permissions/permissions.go
```diff
@@ -0,0 +1,163 @@
...
+
+	if accessorID == "" {
+		return api.Response{}, ErrorPermissions{Wrapped: ErrorMissingAccessorID, Operation: PUT}
+	}
+
+	return c.update(ctx, objectID, accessorType, accessorID, body)
+}
+
+func (c *Client) update(ctx context.Context, objectID string, accessorType string, accessorID string, body []byte) (api.Response, error) {
+	if objectID == "" {
```
Laubi, line 112: These are real Java-style asserts and Go usually does not do that. I would remove them tbh.

@tomazpu, could you maybe bring in a 2nd opinion here (or 3rd :D)

tomazpu (approved):

clients/settings/permissions/permissions_test.go
```diff
@@ -0,0 +1,791 @@
...
+		assert.NotEmpty(t, resp)
+		assert.Equal(t, getResponse, string(resp.Data))
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+	})
+
+	t.Run("If settings object with ID doesn't exists on server returns error", func(t *testing.T) {
+		errorResponse := `{"error": {"code": 404,"message": "Settings not found"}}`
+
+		server := httptest.NewServer(http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+			switch request.Method {
```
Kirdock, line 111: I would like to get rid of all the switch cases, as everything can be done with a server mux. This would also improve readability. No blocker for the PR though

Kirdock (approved):

PR #215, opened 2025-05-06
refactor(segments): remove segment client from `api/clients`

tomazpu:
This PR is removing the segments client from `api/clients` in order to prevent code duplication.

- The segment client in `api/clients` was removed
- Test cases were moved into the `client/segments` (in order to prevent too many changes the current solution with testing was used.)

NONE


clients/segments/segments.go
```diff
@@ -85,46 +77,46 @@ func modifyBody(source []byte) ([]byte, error) {
...
 }
 
-// Get gets a complete configuration of segment with an ID
-func (c Client) Get(ctx context.Context, id string) (Response, error) {
-	resp, err := c.client.Get(ctx, id, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
-	if err != nil {
-		return Response{}, fmt.Errorf(errMsgWithId, "get", id, err)
+func (c Client) Get(ctx context.Context, id string) (api.Response, error) {
+	if id == "" {
+		return api.Response{}, fmt.Errorf(errMsgWithId, "get", id, errors.New(`argument "id" is empty`))
```
Kirdock, line 82: Correct me if I'm wrong, but we decided to go with single quotes, right (also for other usages in here)?

clients/segments/segment_test.go
```diff
@@ -60,22 +57,33 @@ func TestList(t *testing.T) {
...
   ]`
 
-	mockClient := segments.NewMockclient(gomock.NewController(t))
-	mockClient.EXPECT().
-		List(gomock.AssignableToTypeOf(t.Context()), gomock.AssignableToTypeOf(rest.RequestOptions{})).
-		Return(&http.Response{
-			StatusCode: http.StatusOK,
-			Body:       io.NopCloser(strings.NewReader(apiResponse)),
-		}, nil)
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
Kirdock, line 60: [nit]: You could actually split the request and not handle everything in here.
For example, https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/f0021e8adcb7a6d59b98170385bc69ce3c8b45a8/pkg/client/dtclient/settings_client_test.go#L941-L962

clients/segments/segment_test.go
Kirdock, line 1: For some tests, it's hard to see if the correct thing is actually tested.
There are a lot of error checks with different setups, but the outcome is not really validated. testB could actually have the error for testA, but it doesn't check for errorB. Not sure how we want to proceed here

clients/segments/segments.go
tomazpu: True, I just copied as it was. Can do it in the follow-up by simplifying and improving the code. Here, the goal was to only remove the `api/clients`.

clients/segments/segment_test.go
tomazpu: Should be improved, totally agree with you. But in my opinion in the next step.

Kirdock (approved):

clients/segments/segments.go
```diff
@@ -15,49 +15,41 @@
...
-var _ client = (*segments.Client)(nil)
+	ro := rest.RequestOptions{
+		CustomShouldRetryFunc: rest.RetryIfTooManyRequests,
+		QueryParams:           url.Values{"add-fields": []string{"EXTERNALID"}},
+	}
 
-// List gets a complete set of available configs. The Data filed in response is normalized to json list of entries.
-func (c Client) List(ctx context.Context) (Response, error) {
-	resp, err := c.client.List(ctx, rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
+	resp, err := c.restClient.GET(ctx, path, ro)
```
arthurpitman, line 50: Nit: is there a stylistic reason to have `ro` here?

clients/segments/segments.go
```diff
@@ -133,44 +125,72 @@ func (c Client) Update(ctx context.Context, id string, data []byte) (Response, e
...
+	path, err := url.JoinPath(endpointPath, id)
 	if err != nil {
-		return Response{}, fmt.Errorf(errMsgWithId, "update", id, err)
+		return api.Response{}, fmt.Errorf(errMsgWithId, "update", id, err)
 	}
 
-	updateResourceResp, err := c.client.Update(ctx, id, data, rest.RequestOptions{
+	resp, err := c.restClient.PUT(ctx, path, bytes.NewReader(body), rest.RequestOptions{
 		CustomShouldRetryFunc: rest.RetryIfTooManyRequests,
 		QueryParams:           map[string][]string{"optimistic-locking-version": {fmt.Sprint(getResponse.Version)}}})
```
arthurpitman, line 145: Would this work? I had to do a double-take at `fmt.Sprint(...)`...
```suggestion
		QueryParams:           map[string][]string{"optimistic-locking-version": {strconv.Itoa(getResponse.Version)}}})
```

arthurpitman (approved):

PR #216, opened 2025-05-06
feat: support custom retry delay and amount for rest client calls

Kirdock:
Specific clients in Monaco CLI, like settings and config clients, have different retry logic. If the core retry logic should be used, the retry options for CRUD calls need to be extended. Therefore, CRUD requests now accept a custom delay and max retry amount, which is preferred over the client-set one.

Should we create a follow-up to move the request options into a struct? Note that the `RequestOptions` struct can't be reused because we must differentiate between empty value and nil (for example, zero retries or zero delay). Changing it to a struct also means a breaking change for the Monaco CLI (not for Terraform, as this is not used on their side).

- No breaking change for consumers of the lib
- Retry delay and max retry amount can now be set on CRUD operations

### DIFF
--- a/api/rest/client_test.go
+++ b/api/rest/client_test.go
@@ -27,7 +27,9 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/internal/pointer"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/testutils"
 )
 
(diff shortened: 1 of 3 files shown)


api/rest/client_test.go
```diff
@@ -328,6 +329,30 @@ func TestClient_WithRetries(t *testing.T) {
...
+		},
+	}))
+
+	ctx := testutils.ContextWithLogger(t)
+
+	resp, err := client.GET(ctx, "", RequestOptions{MaxRetries: testutils.Pointer(1)})
+	require.NoError(t, err)
+	assert.Equal(t, http.StatusTooManyRequests, resp.StatusCode)
+	assert.Equal(t, 2, apiHits)
+}
```
arthurpitman, line 354: Shall we add a test with a custom `DelayAfterRequest`?

arthurpitman (commented):
Follow up idea: use `testutils.Pointer` in the Monaco CLI and then remove` internal/pointer/pointer.go` there

api/rest/client_test.go
Kirdock: The thing is, how would you test that? I could test the merge function instead. This one is private, but in the same package

Kirdock:
I've put the pointer utils now into `internal/pointer` 👍

api/rest/client_test.go
Kirdock: I was curious about how to test it, so I added one anyway 😅

arthurpitman (approved):

PR #221, opened 2025-05-23
feat(clients): Allows settings custom http headers for client factory

d0weinberger:
#### **Why** this PR?
We want to have the option in Monaco to add  custom HTTP headers to requests.

#### **What** has changed?
Adds a map of customHeaders to the factory struct. Adds `WithCustomHeaders` function. 

#### **How** does it do it?
Sets headers, when creating REST clients.

#### How is it **tested**?
No tests added yet. I'd like to test headers set for created REST client but they are private. I don't want to expose them just for the test. Do you have ideas?

Allows consumers of the library to add their custom HTTP headers to requests.

### DIFF
--- a/clients/factory.go
+++ b/clients/factory.go
@@ -67,6 +67,7 @@ type factory struct {
 	concurrentRequestLimit int                       // The number of allowed concurrent requests
 	rateLimiterEnabled     bool                      // Enables rate limiter for clients
 	retryOptions           *rest.RetryOptions        // The retry strategy
+	customHeaders          map[string]string         // Custom HTTP headers
 }
 
 // WithOAuthCredentials sets the OAuth2 client credentials configuration for the factory.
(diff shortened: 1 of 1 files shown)


clients/factory.go
```diff
@@ -250,6 +257,9 @@ func (f factory) createRestClient(u string, httpClient *http.Client) (*rest.Clie
 	if f.userAgent != "" {
 		restClient.SetHeader("User-Agent", f.userAgent)
 	}
+	for headerKey, headerValue := range f.customHeaders {
```
Laubi, line 260: This might overwrite the `User-Agent` above, we should add this into the documentation on L136

Kirdock (approved):

Laubi (approved):

clients/factory.go
arthurpitman: Also, due to the way the rest client (currently) works, these also could potentially overwrite any headers set on the request, such as content type.

PR #224, opened 2025-06-04
refactor(automation): fix Sonar and error message issues

Kirdock:
#### **Why** this PR?
There are some sonar issues coming from https://github.com/Dynatrace/dynatrace-configuration-as-code-core/pull/223. This PR fixes those and aligns all error messages that are returned.

#### **What** has changed?
- Function complexity reduced
- Duplications reduced
- Adapted interfaces
- Aligned error messages

#### **How** does it do it?
- listResponse correctly represents the response now
- split List into another listPage, to reduce complexity
- moved some errors into constants and wrappers
- by aligning error messages (same structure)

#### How is it **tested**?
Current tests already cover it.

Doesn't affect users.

### Notes to the reviewer
We could handle the `wfAdminAccess` differently, by returning a custom error and checking this one in the List function (`if errors.As(err, &ErrWfAdminForbidden)` then set `wfAdmin` to `false`. How about this solution?

### DIFF
--- a/clients/automation/automation.go
+++ b/clients/automation/automation.go
@@ -46,16 +46,24 @@ const (
 	SchedulingRules
 )
 
-var resources = map[ResourceType]Resource{
-	Workflows:         {Path: "/platform/automation/v1/workflows"},
-	BusinessCalendars: {Path: "/platform/automation/v1/business-calendars"},
-	SchedulingRules:   {Path: "/platform/automation/v1/scheduling-rules"},
-}
+var (
+	resources = map[ResourceType]Resource{
+		Workflows:         {Path: "/platform/automation/v1/workflows"},
+		BusinessCalendars: {Path: "/platform/automation/v1/business-calendars"},
+		SchedulingRules:   {Path: "/platform/automation/v1/scheduling-rules"},
+	}
+	ErrMissingID = errors.New("id must be non empty")
+)
 
 type listResponse struct {
-	api.Response
 	Count   int               `json:"count"`
 	Objects []json.RawMessage `json:"results"`
+}
+
+type listNextResult struct {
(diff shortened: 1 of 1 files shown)


tomazpu (dismissed):

clients/automation/automation.go
```diff
@@ -195,69 +213,85 @@ func (a Client) List(ctx context.Context, resourceType ResourceType) (api.PagedL
...
+	return retVal, nil
+}
 
-		// if Workflow API rejected the initial request with admin permissions -> retry without
-		if wfAdminAccess && resp.StatusCode == http.StatusForbidden {
-			wfAdminAccess = false
-			resp.Body.Close()
-			continue
-		}
+func (a Client) listPage(ctx context.Context, resourceType ResourceType, wfAdminAccess bool, retrieved int) (listNextResult, error) {
```
arthurpitman, line 239: From the perspective of this function, I think calling it `offset` might be better.

```suggestion
func (a Client) listPage(ctx context.Context, resourceType ResourceType, wfAdminAccess bool, offset int) (listNextResult, error) {
```

clients/automation/automation.go
```diff
@@ -46,18 +46,36 @@ const (
...
+	errMsgWithId    = "failed to %s automation resource of type %v with id %s: %w"
+	getOperation    = "get"
+	updateOperation = "update"
+	listOperation   = "list"
+	deleteOperation = "delete"
+	createOperation = "create"
+)
+
+var (
+	resources = map[ResourceType]Resource{
```
arthurpitman, line 60: As we move towards changing what a resource is, what is the feeling about splitting this into three separate resource in a future follow-up?

PR #226, opened 2025-06-04
refactor(buckets): remove bucket client from `api/clients`

Kirdock:
#### **Why** this PR?

#### **What** has changed?
The bucket API client is removed and merged with `clients/bucket`.

#### How is it **tested**?
Existing tests cover the changes.

Consumers of the bucket API client need to migrate to the client/bucket one.


clients/buckets/bucket.go
```diff
@@ -369,7 +381,7 @@ func (c Client) upsert(ctx context.Context, bucketName string, data []byte) (Res
 	if b, err := unmarshalJSON(apiErr.Body); err != nil && b.Status == "deleting" {
 		logger.V(1).Info(fmt.Sprintf("Bucket '%s' is being deleted. Waiting before re-creation...", b.BucketName))
 		if err := c.awaitBucketRemoved(ctx, bucketName); err != nil {
-			return Response{}, err
+			return api.Response{}, err
 		}
 		return c.Create(ctx, bucketName, data)
 	}
```
Kirdock, line 387: Is this dead code? First, it should be `err == nil`, and second, the response doesn't contain the bucket status.
This is the response I get while the bucket is being deleted.
image

Kirdock, line 387: This is dead code, right? First, it should be `err == nil`, and second, If there is an error during create (conflict), this will lead to this error
image

Kirdock: Removed it

clients/buckets/bucket.go
```diff
@@ -365,15 +377,6 @@ func (c Client) upsert(ctx context.Context, bucketName string, data []byte) (Res
 		return api.Response{}, apiErr
 	}
 
-	// If bucket is currently being deleted, wait for it to be gone, then re-create it
```
d0weinberger, line 368: Is this block moved to somewhere else or not needed anymore? I didn't find it

clients/buckets/bucket.go
```diff
@@ -393,7 +396,7 @@ func (c Client) create(ctx context.Context, bucketName string, data []byte) (api
 	if err := setBucketName(bucketName, &data); err != nil {
 		return api.Response{}, fmt.Errorf("unable to set bucket name: %w", err)
 	}
-	r, err := c.apiClient.Create(ctx, data)
+	r, err := c.restClient.POST(ctx, endpointPath, bytes.NewReader(data), rest.RequestOptions{CustomShouldRetryFunc: rest.RetryIfTooManyRequests})
 	if err != nil {
 		return api.Response{}, fmt.Errorf("failed to create bucket with name '%s': %w", bucketName, err)
```
d0weinberger, line 401: Also a difference in the error message, but here I think it's good that we lose the nested `"failed to create new bucket: %w"`

clients/buckets/bucket.go
```diff
@@ -318,8 +321,17 @@ func (c Client) Upsert(ctx context.Context, bucketName string, data []byte) (Res
...
 //   - error: An error if the HTTP call fails or another error happened.
-func (c Client) Delete(ctx context.Context, bucketName string) (Response, error) {
-	resp, err := c.apiClient.Delete(ctx, bucketName)
+func (c Client) Delete(ctx context.Context, bucketName string) (api.Response, error) {
+	if bucketName == "" {
+		return api.Response{}, ErrBucketEmpty
+	}
+	path, err := url.JoinPath(endpointPath, bucketName)
+	if err != nil {
+		return api.Response{}, fmt.Errorf("failed to create URL: %w", err)
```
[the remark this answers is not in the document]
Kirdock: The funny thing is that we had both variants before (Get without some context, Delete with some context).
I'll restructure all the error messages here a bit

clients/buckets/bucket.go
Kirdock: Yes, commented it. It was dead code, so I removed it

clients/buckets/bucket.go
Kirdock: I've aligned all the error messages to match the slo and segments. Not like "permissions" because that would probably open another pandora's box 😅

tomazpu (approved):

tomazpu:
I have points regarding the error handling and messages, but i would see it as a follow-up.

d0weinberger (approved):

PR #228, opened 2025-06-12
feat: Allow creation of platform clients using platform tokens

Laubi:
#### **Why** this PR?
With this changeset, library users are able to create platform clients using a platform token.
If both options are set, the platform token takes precedence. 

#### **What** has changed?
1. Refactored the Auth Transport to use the OAuth StaticSource for simplicity
2. Added tests for the creation of basic clients
3. The factory can now create platform clients based on platform tokens
4. Added tests for the factory to ensure that the correct authentication is used 

#### How is it **tested**?
1. Proper tests that the OAuth token and platform token is set for the requests

Users of the library may now use platform tokens to create platform clients.

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -4,7 +4,7 @@ sonar.projectName=dynatrace-configuration-as-code-core
 
 sonar.language=go
 sonar.sources=.
-sonar.exclusions=**/*_test.go, testutils/**, gen/**, .github/*, tools/*, lint-report.xml
+sonar.exclusions=**/*_test.go, testutils/**, gen/**, .github/*, tools/*, lint-report.xml, **/testutils/**
 
 sonar.tests=.
 sonar.test.inclusions=**/*_test.go


clients/factory.go
```diff
@@ -244,7 +262,7 @@ func (f factory) CreateClassicClient() (*rest.Client, error) {
 		return nil, ErrClassicURLMissing
 	}
 
-	return f.createRestClient(f.classicURL, auth.NewTokenBasedClient(f.accessToken))
+	return f.createRestClient(f.classicURL, auth.NewApiTokenBasedClient(context.TODO(), f.accessToken))
```
Kirdock, line 265: Is there a follow-up for this one or should we now add the context to the parameters?

Laubi: Adding it is a breaking change, that's why I didn't do it. 
There are for sure possible follow ups, but it's not urgent at all. In fact, the context is not used at all, `nil` would have been a valid value. But sonar shouted at me :)

Kirdock (approved):

api/auth/auth.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+func NewTokenBasedClient(ctx context.Context, apiToken string) *http.Client {
+	return oauth2.NewClient(ctx, oauth2.StaticTokenSource(&oauth2.Token{TokenType: "Api-Token", AccessToken: apiToken}))
+}
+
+func NewPlatformTokenClient(ctx context.Context, platformToken string) *http.Client {
```
arthurpitman, line 31: Nit: either we should add "Based" to this function name or remove it from the other two.

arthurpitman (approved):

PR #230, opened 2025-06-12
test: Introduce end-to-end tests

Laubi:
#### **Why** this PR?
We currently do not have any tests that test the 'real' behavior of the core library. We would not notice that an authentication method is broken until a consumer uses the library and notices it. 

This PR changes that. We introduce e2e tests that are executed
* nightly
* on `main`
* on branches with the `run-e2e-test` label

I suggest the following rules:
* No e2e test folder - put the e2e tests in the same directory as the functionality.
* The `e2e` tests must include the `e2e` built tag

#### 🚨 Security considerations

Workflows with the `pull_request_target` are dangerous to use. 
They are executed in the context of the target repository; thus, an attacker could create a PR to read all secrets in our repository.
We mitigate this issue by requiring the `run-e2e-test` label.
This label can only be set by maintainers and must only be set on reviewed PRs.

I reused the same mechanism as we have in the [Monaco repository](https://github.com/Dynatrace/dynatrace-configuration-as-code).

Until this PR is discussed, no secrets will be set in this repository - just to be sure.

#### **What** has changed?
We now have e2e tests 🎉 

#### **How** does it do it?
* Secrets are passed in the workflow
* Secrets can be used in the tests

**Note**: Secrets are not yet defined in GitHub to first review for potential issues.

#### How is it **tested**?
It's about tests!

Users of the library can be more sure that it works 🎉

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -1,0 +1,47 @@
+name: End-to-end tests
+on:
+  push:
+    branches:
+      - main
+  schedule:
+   - cron: '0 0 * * *' # nightly build ensure E2E tests run daily and catch any breaking API changes
+  pull_request_target:
+    types:
+      - labeled
+
+jobs:
+  integration-test:
+    name: 🌍 Integration tests
+    if: github.event.action != 'labeled' || github.event.label.name == 'run-e2e-test'
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


tomazpu:
Looks good to me, as in the description, the attack vector is known and handled with a label.

tomazpu (approved):

PR #231, opened 2025-06-17
refactor(errors): Introduction of `errors` package and example usage in the `segments` client

tomazpu:
#### **Why** this PR?
To return better errors from the library, I am introducing the errors package with the first three types of errors returned from a client.

#### **What** has changed?

- Introduction of errors package
- First three base error types introduced (`ErrorValidation`, `ErrorClient`, and `ErrorRuntime`)
- Example usage in the segments client

#### How is it **tested**?

- Existing test's where enhanced

- Improved Errors are returned for customers of the lib.

### DIFF
--- a/clients/segments/segments.go
+++ b/clients/segments/segments.go
@@ -32,6 +32,7 @@ const endpointPath = "platform/storage/filter-segments/v1/filter-segments"
 const resource = "segments"
 
 var basePath = url.URL{Path: endpointPath}
+var idValidationErr = api.ValidationError{Field: "id", Reason: "is empty"}
 
 func NewClient(client *rest.Client) *Client {
 	return &Client{restClient: client}
@@ -79,7 +80,7 @@ func modifyBody(source []byte) ([]byte, error) {
 
 func (c Client) Get(ctx context.Context, id string) (api.Response, error) {
 	if id == "" {
-		return api.Response{}, api.ValidationError{Field: "id", Reason: "is empty"}
+		return api.Response{}, idValidationErr
 	}
 
 	path := basePath.JoinPath(id).String()
@@ -105,7 +106,7 @@ func (c Client) Create(ctx context.Context, body []byte) (api.Response, error) {
 
 func (c Client) Update(ctx context.Context, id string, body []byte) (api.Response, error) {
 	if id == "" {
-		return api.Response{}, api.ValidationError{Field: "id", Reason: "is empty"}
+		return api.Response{}, idValidationErr
 	}
 
 	existing, err := c.Get(ctx, id)
@@ -146,7 +147,7 @@ func (c Client) Update(ctx context.Context, id string, body []byte) (api.Respons
 
 func (c Client) Delete(ctx context.Context, id string) (api.Response, error) {
 	if id == "" {
-		return api.Response{}, api.ValidationError{Field: "id", Reason: "is empty"}
+		return api.Response{}, idValidationErr
 	}
 
 	path := basePath.JoinPath(id).String()


internal/errors/errors.go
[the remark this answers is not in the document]
tomazpu: Very good point, thanks!

clients/segments/segments.go
```diff
@@ -178,14 +164,14 @@ func (c Client) GetAll(ctx context.Context) ([]api.Response, error) {
 		Uid string `json:"uid"`
 	}
 	if err = json.Unmarshal(listResp.Data, &segments); err != nil {
-		return nil, fmt.Errorf(errMsg, "get all", err)
+		return nil, intErr.ErrorRuntime{Resource: resource, Reason: "unmarshalling failed", Wrapped: err}
```
Kirdock, line 167: The `ErrorRuntime` error is always used when an error occurs, right? Would the reason here not be duplicate information compared to the returned error?

internal/errors/errors.go
```diff
@@ -0,0 +1,96 @@
...
+	// Wrapped contained the raw error returned from the client
+	Wrapped error
+
+	// Identifier being not empty the error message will contain the value set
+	Identifier string
+
+	// Operation is the HTTP method that was used by the client
+	Operation string
+
+	// Resource is a string of the resource
```
Laubi, line 32: Please add here what a resource is. 
`Resource is a string of the resource` does not tell me anything new that `Resource string` didn't tell me before 😅 

Same with `Identifier`. 

`Operation` -> is this the method (`GET/POST/PUT`) or something more rest friendly aka `List/Update/Create`)?

internal/errors/errors.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	return fmt.Sprintf("failed to %s %s: %v", e.Operation, e.Resource, e.Wrapped)
+}
+
+func (e ErrorClient) Unwrap() error {
+	return e.Wrapped
+}
+
+type ErrorValidation struct {
```
Laubi, line 51: We should add when what error should be used. `ErrorValidation` for example is meaningless. What validation? Server side? Client Side? Why set a field? Is this something that can be reacted to? Why is it not `ErrorClient`?

clients/segments/segment_test.go
```diff
@@ -75,16 +76,17 @@ func TestList(t *testing.T) {
 }
 
 func TestGet(t *testing.T) {
-	t.Run("when called without id parameter, returns an error", func(t *testing.T) {
+	t.Run("when called without id parameter, returns an validation error", func(t *testing.T) {
```
Laubi, line 79:
```suggestion
	t.Run("when called without id parameter, returns a validation error", func(t *testing.T) {
```

clients/segments/segments.go
```diff
@@ -125,44 +119,36 @@ func (c Client) Update(ctx context.Context, id string, body []byte) (api.Respons
 	}
 	err = json.Unmarshal(existing.Data, &getResponse)
 	if err != nil || getResponse.Version == 0 || getResponse.Owner == "" {
-		return api.Response{}, fmt.Errorf(errMsgWithId, "update", id, err)
+		return api.Response{}, intErr.ErrorValidation{Field: "version, owner", Reason: "invalid"}
```
d0weinberger, line 122: Here it gets tricky, because we would actually need to transport the information that version OR owner is invalid (as opposed to, maybe, we just fail when BOTH are invalid). Maybe add it to the reason? Instead of "invalid" say "at least one is invalid"?

clients/segments/segments.go
arthurpitman, line 167: What do you think about using an `UnmashalingError` and then wrapping it with an `ErrorRuntime ?

internal/errors/errors.go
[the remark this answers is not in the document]
Laubi: I would vote not having an errors package at all. Isn't there some 'api' or idk package that we can move those errors to? 🤔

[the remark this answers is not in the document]
tomazpu: I'll follow the pattern of the standard lib, where packages have `error.go` files that export error structs in their package.

internal/errors/errors.go
```diff
@@ -0,0 +1,96 @@
...
+	// Wrapped contained the raw error
+	Wrapped error
+
+	// Field represents the name of the field that the validation failed
+	Field string
+
+	// Reason contains why the validation failed
+	Reason string
```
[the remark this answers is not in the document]
tomazpu: Added context on the idea of what it should provide. Please recheck if that makes sense for you.

d0weinberger (approved):

PR #237, opened 2025-07-08
refactor(document): merge clients

Kirdock:
#### **Why** this PR?

#### **What** has changed?
The document API client is removed and merged with `clients/document`.
Additional refactoring:
- Responses and status checks are now done via `NewResponseFromHTTPResponse`.
- Instead of two approaches on how to extract the content and the metadata, there is now one. This also removes the multipart files.
- Aligned error messages. A proper error message is attached when an error is returned.
- Reduced some functions' complexities by introducing new functions

#### How is it **tested**?
Existing tests cover the changes.
Updated the client in Monaco and ran E2E. No errors 👍 
https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1976

Consumers of the document API client need to migrate to the client/document one.

**Issue**: CA-15353

## Test coverage
Regarding test coverage, there are two problems
1. It's not possible to fail `url.JoinPath`. It can only error through the provided base URL, and ours is static
2. Client errors. At the moment, our `testutils.ResponseDef` utility function doesn't support returning errors. In order to make this work, stubs need to be introduced, which would have been a bit overkill for this task, as it would also involve re-structuring all the other tests of the document client. Could do a follow-up, though.


clients/documents/documents.go
```diff
@@ -292,73 +263,97 @@ func (c Client) patchWithRetry(ctx context.Context, id string, version int, d do
 	return
 }
 
-func (c Client) patch(ctx context.Context, id string, version int, d documents.Document) (api.Response, error) {
-	resp, err := processHttpResponse(c.client.Patch(ctx, id, version, d))
+func (c Client) patch(ctx context.Context, id string, version int, d Document) (api.Response, error) {
+	if id == "" {
+		return api.Response{}, ErrIDEmpty
+	}
```
arthurpitman, line 269: I'm not sure that we really need this check, but if we keep it perhaps we should wrap it to indicate that it is coming from the patch?

Kirdock: I also thought about that, and I wasn't sure. The only possible way that this condition becomes true is when the POST inside Create returns success, and the body doesn't contain the ID.
I can remove it 👍

api/clients/documents/documents.go
```diff
@@ -1,161 +0,0 @@
...
-
-	resp, err := c.client.GET(ctx, path, rest.RequestOptions{})
-	if err != nil {
-		return nil, fmt.Errorf("unable to get object with ID %s: %w", id, err)
-	}
-
-	return resp, err
-}
-
-// Create posts the content of the document to the server using http POST to create new document. NOTE: some of the arguments of the document are ignored due the design of the HTTP API.
```
d0weinberger, line 65: Would it make sense to preserve some of these comments? They are now also removed together with the client. But maybe you could copy them to the merged client so we don't lose some explanatory comments.

clients/documents/documents.go
```diff
@@ -292,73 +267,90 @@ func (c Client) patchWithRetry(ctx context.Context, id string, version int, d do
...
-	r := multipart.NewReader(resp.Body, boundary)
-	for p, err := r.NextPart(); err != io.EOF; p, err = r.NextPart() {
-		if err != nil {
-			return api.MultipartResponse{}, fmt.Errorf("failed to read the content of the document resource with id %s: %w", id, err)
-		}
-		buf, err := io.ReadAll(p)
-		if err != nil {
-			return api.MultipartResponse{}, fmt.Errorf("failed to read the content of the document resource with id %s: %w", id, err)
-		}
+	reader := multipart.NewReader(bytes.NewReader(resp.Data), boundary)
```
d0weinberger, line 324: I think it would still make sense to split functionality concerning multipart reading away, to a different file and a different component. It makes it easier to test and decreases complexity of at least this function.

api/clients/documents/documents.go
d0weinberger: Yeah, on a second read, I think you're right 😅

PR #238, opened 2025-07-10
refactor(openpipeline): merge clients

Kirdock:
#### **Why** this PR?

#### **What** has changed?
The openpipeline API client is removed and merged with clients/openpipeline.
Also removes some logs. The error already contains the needed information.

#### How is it **tested**?
Existing tests cover it, and new ones are added.

Doesn't affect users.

**Issue:** CA-15354


clients/openpipeline/openpipeline.go
```diff
@@ -15,24 +15,31 @@
...
 )
 
-const bodyReadErrMsg = "unable to read API response body"
-const maxUpdateAttempts = 10
+const (
+	maxUpdateAttempts        = 10
+	openPipelineResourcePath = "/platform/openpipeline/v1/configurations"
+
+	errMsg       = "failed to %s openpipeline resource: %w"
+	errMsgWithId = "failed to %s openpipeline resource with id %s: %w"
```
d0weinberger, line 35: Is it `type id` or was that an error in the previous log line on line 58?

clients/openpipeline/openpipeline.go
```diff
@@ -41,55 +48,60 @@ type ListResponse struct {
...
+		return nil, fmt.Errorf(errMsg, listOperation, err)
 	}
 
 	var resources []ListResponse
-	err = json.Unmarshal(body, &resources)
+	err = json.Unmarshal(resp.Data, &resources)
 	if err != nil {
-		logr.FromContextOrDiscard(ctx).Error(err, "failed to unmarshal json response")
-		return nil, api.NewAPIErrorFromResponseAndBody(resp, body)
+		return nil, fmt.Errorf(errMsg, listOperation, err)
```
d0weinberger, line 93: We don't return an `APIError` anymore then. Will this cause problems; to we assert on `APIError` for openpipeline somewhere?

Kirdock: I would not see this as a problem because the error is not related to an invalid status code. It's not an API error but a validation error I would say

Kirdock: To make it more weird: Before it was an APIError with a status code 200 😅

clients/openpipeline/openpipeline.go
```diff
@@ -41,55 +48,60 @@ type ListResponse struct {
...
 
 	return api.NewResponseFromHTTPResponse(resp)
 }
 
 func (c Client) List(ctx context.Context) ([]ListResponse, error) {
-	resp, err := c.client.List(ctx, rest.RequestOptions{})
+	httpResp, err := c.restClient.GET(ctx, openPipelineResourcePath, rest.RequestOptions{})
 	if err != nil {
-		return nil, fmt.Errorf("failed to list openpipeline resources: %w", err)
+		return nil, fmt.Errorf(errMsg, listOperation, err)
```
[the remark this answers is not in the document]
Kirdock: I kept it the same as for segments and slo-v2. That's why I chose singular

tomazpu (approved):

PR #240, opened 2025-07-22
feat(settings): support adminAccess query param

Kirdock:
#### **Why** this PR?
The `adminAccess` query parameter was added to support fetching permissions of different owners. When it is set to `true` and permission is given to `settings:objects:admin`, the requester is seen as the owner of the setting.

#### **What** has changed?
A new `adminAccess` parameter is added to all permission calls.

#### How is it **tested**?
Existing tests are adapted to check for the existence of the `adminAccess` query parameter.

The parameters for permission calls have changed. Now the `adminAccess` query parameter has to be provided.

**Issue:** CA-15757

### DIFF
--- a/clients/settings/permissions/permissions.go
+++ b/clients/settings/permissions/permissions.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"net/url"
+	"strconv"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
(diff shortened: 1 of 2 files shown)


d0weinberger (dismissed):

clients/settings/permissions/permissions_test.go
arthurpitman, line 1: Do we still have test cases where `adminAccess` is not set to `true`?

arthurpitman: Fair point, I figured yes because of `getDefaultRequestOptions(...)`

clients/settings/permissions/permissions.go
```diff
@@ -153,11 +154,18 @@ func (c *Client) delete(ctx context.Context, objectID string, accessorType strin
...
+	httpResp, err := c.client.DELETE(ctx, path, getDefaultRequestOptions(adminAccess))
 
 	if err != nil {
 		return api.Response{}, ErrorPermissions{Wrapped: err, Operation: DELETE}
 	}
 
 	return api.NewResponseFromHTTPResponse(httpResp)
 }
+
+func getDefaultRequestOptions(adminAccess bool) rest.RequestOptions {
```
Laubi, line 166: `getDefault` would mean that there are other options. Wouldn't just `getRequestOptions` be enough?

clients/settings/permissions/permissions.go
```diff
@@ -153,11 +154,18 @@ func (c *Client) delete(ctx context.Context, objectID string, accessorType strin
...
+	return rest.RequestOptions{
+		CustomShouldRetryFunc: rest.RetryIfTooManyRequests,
+		QueryParams:           url.Values{"adminAccess": []string{strconv.FormatBool(adminAccess)}},
```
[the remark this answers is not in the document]
Kirdock: I would not see the need to add additional complexity to the code just for "it could be removed.". To keep it simple, I would still stick to this solution if there isn't a major problem with it.

clients/settings/permissions/permissions.go
Kirdock: Good point. My original thought was that if there is some endpoint that needs different ones, this would be the default case, and other calls can adapt based on that. Because we don't have any other options here, `getRequestOptions` is sufficient, and we can adapt it in the future if even needed.

clients/settings/permissions/permissions.go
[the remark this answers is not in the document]
Laubi: I won't force any change now, however, I still stand by 'don't use (send) what you don't need' :D

PR #251, opened 2025-09-01
feat: forward context to classic client creation

Kirdock:
The context was not forwarded to the creation of the oauth2 client, which prevented us from overriding the client used in the oauth2 lib

#### **Why** this PR?
For some configurations, we need to override the HTTP client used. This is only possible by adding the custom client to the context, which is then later on used by the OAuth2 lib.

#### **What** has changed?
A new function was added that accepts a context when creating a classic client.
The old function wasn't touched as this would result in a breaking change and several adjustments in the Terraform provider.

#### **How** does it do it?
By adding a new function that accepts and forwards the context.

#### How is it **tested**?
New tests added for creating the client

It doesn't affect users but consumers of the lib. They can now override the http client by setting it in the context
```go
context.WithValue(ctx, oauth2.HTTPClient, &http.Client{
	Transport: &http.Transport{
		...
	},
	...
})
```

### DIFF
--- a/clients/factory.go
+++ b/clients/factory.go
@@ -243,6 +243,11 @@ func (f factory) CreatePlatformClient(ctx context.Context) (*rest.Client, error)
 
 // CreateClassicClient creates a REST client configured for accessing classic APIs.
 func (f factory) CreateClassicClient() (*rest.Client, error) {
+	return f.CreateClassicClientWithContext(context.TODO())
+}
+
+// CreateClassicClientWithContext creates a REST client configured for accessing classic APIs with a given context.
+func (f factory) CreateClassicClientWithContext(ctx context.Context) (*rest.Client, error) {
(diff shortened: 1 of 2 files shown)


d0weinberger (dismissed):

clients/factory.go
```diff
@@ -254,6 +254,19 @@ func (f factory) CreateClassicClient() (*rest.Client, error) {
 	return f.createRestClient(f.classicURL, auth.NewAPITokenClient(context.TODO(), f.accessToken))
 }
 
+// CreateClassicClientWithContext creates a REST client configured for accessing classic APIs with a given context.
```
arthurpitman, line 257: For the implementation of `CreateClassicClient(...)`, what about just calling this method with a `context.TODO()` to prevent having to duplicate the checks?

PR #257, opened 2025-09-22
feat(document): Support `originExtensionId` in metadata

arthurpitman:
#### **Why** this PR?
This PR adds `originExtensionId` to `documents.Metadata`. This field, together with `originAppId` allows ready-made documents to be identified, which cannot be modified.

#### **What** has changed and **how** does it do it?
-The `documents.Metadata` struct is updated.
-The `documents.List(...)` method now requests the `originExtensionId` field be included in results. No change is required for `documents.Get(...)` as `originExtensionId` is always returned in the GET responses.

#### How is it **tested**?
Existing tests are updated.

None.

**Issue:** CA-16532

### DIFF
--- a/clients/documents/documents.go
+++ b/clients/documents/documents.go
@@ -138,7 +138,7 @@ func (c Client) List(ctx context.Context, filter string) (ListResponse, error) {
 
 	for result.NextPageKey != nil {
 
-		queryParams := url.Values{"filter": {filter}}
+		queryParams := url.Values{"filter": {filter}, "add-field": {"originExtensionId"}}
 		if *result.NextPageKey != "" {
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
 	github.com/stretchr/testify v1.11.1
+	go.uber.org/mock v0.6.0
```
Kirdock, line 10: I assume this addition and the `go.sum` one were unintended? I don't see any usages

arthurpitman: Found the culprit, an outdated `segments_mock.go` 👍

Kirdock (approved):

PR #301, opened 2026-04-02
feat: add extension client

Kirdock:
#### **Why** this PR?
To support the new platform extension endpoint

#### **What** has changed?
This adds an extension client, which supports CRUD operations for monitoring configurations and GET for extensions

#### How is it **tested**?
Unit tests added
Already tested locally in Terraform

They can create, update, delete extension monitoring configurations and get a list of all installed extensions.

### DIFF
--- a/clients/extensions/client_test.go
+++ b/clients/extensions/client_test.go
@@ -1,0 +1,596 @@
+// @license
+// Copyright 2026 Dynatrace LLC
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package extensions_test
+
+import (
+	"io"
+	"net/http"
+	"testing"
+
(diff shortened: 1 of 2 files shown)


clients/extensions/client_test.go
```diff
@@ -0,0 +1,596 @@
...
+
+	t.Run("errors if called without configuration ID", func(t *testing.T) {
+		client := extensions.NewClient(&rest.Client{})
+
+		err := client.DeleteMonitoringConfiguration(t.Context(), "com.dynatrace.extension.foo", "")
+
+		assert.ErrorIs(t, err, api.ValidationError{Resource: "monitoring-configurations", Field: "configuration-id", Reason: "is empty"})
+	})
+
+	t.Run("errors if monitoring configuration with ID doesn't exist on server", func(t *testing.T) {
```
arthurpitman, line 558: I don't think adding another test for deleting with an extension name that doesn't exist would bring any benefit as the logic is the same, just a slightly different error message, right?

arthurpitman (approved):
This looks good, and I think it is the most uniform client so far 👍 

Just one question: do we need a way to manage the active version, i.e. the `/extensions/{extension-name}/environment-configuration` endpoint? Or is that something for another PR?

Kirdock:
> This looks good, and I think it is the most uniform client so far 👍
> 
> Just one question: do we need a way to manage the active version, i.e. the `/extensions/{extension-name}/environment-configuration` endpoint? Or is that something for another PR?

Active version and installing/uninstalling extensions is out of scope 👍

clients/extensions/client_test.go
Kirdock: Yes, it's just a different response body from the API

PR #312, opened 2026-05-26
fix: Clients should not customize retry options

arthurpitman:
#### **Why** this PR?
- Workflows retried all admin-mode requests unless a 'forbidden' response was received. As the API checks other things first, this led to a large number of retries if, for example, an invalid ID was used.
- Many API clients were customizing retry behavior with the generic `RetryIfTooManyRequests`. Instead, they should honor the per-client retry settings.
- In general, it makes sense to specify where clients _should_ retry, rather than retrying everything with exceptions. 

#### **What** has changed an **how** does it do it?
- The automation client no longer provides a custom retry function for workflow requests with admin access, and its Delete method no longer wraps retry logic around the admin-access options.
- Automation and document delete methods no longer retry everything with the exception of not found.
- All per-request CustomShouldRetryFunc overrides have been removed from the buckets, segments, slo, documents, and settings/permissions clients. These clients now rely on the ShouldRetryFunc configured at the REST client level.
- RetryIfTooManyRequests has been renamed to RetryIfTooManyRequestsOrServiceUnavailable and now also retries on HTTP 503, matching what most callers actually need.

#### How is it **tested**?
- A new test case "GET - Invalid ID does not retry" is added in clients/automation/automation_test.go and verifies that no retries occur for a non-retryable status code.

- Library users should explicitly set client-level retry options, most likely to `RetryIfTooManyRequestsOrServiceUnavailable`.
- `RetryIfTooManyRequests` has been renamed to `RetryIfTooManyRequestsOrServiceUnavailable`.

### DIFF
--- a/clients/automation/automation.go
+++ b/clients/automation/automation.go
@@ -306,9 +306,6 @@ func (a Client) makeRequestWithAdminAccess(resourceType ResourceType, request fu
 	if resourceType == Workflows {
 		opts := rest.RequestOptions{
 			QueryParams: url.Values{"adminAccess": []string{"true"}},
-			CustomShouldRetryFunc: func(resp *http.Response) bool {
-				return rest.RetryIfNotSuccess(resp) && (resp.StatusCode != http.StatusForbidden)
-			},
 		}
 		resp, err := request(opts)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


clients/automation/automation.go
Laubi, line 164: [just a note] I do not like that we close the Body in a followup call, as it's really always a hassle to double check if we do it correctly. We should eventually think about if we can do it better 🤔

Laubi (approved):

clients/automation/automation.go
arthurpitman: For the moment, I have marked the SonarCloud issues as false positives, but yes, we should come back to this.

